PR #10, opened 2022-10-27
Refactor navigation within /write into seperate component

marcel-klasse:
Create new component for navigation within services. 
In preperation for more services to follow.

image


components/UI/ServiceSubmenu.js
```diff
@@ -0,0 +1,76 @@
...
+    return (<Header>
+        <h1>{ title }</h1>
+        { React.Children.map(children, child =>
+            React.cloneElement(child, { opensubmenu, setOpenSubmenu }),
+        ) }
+    </Header>
+    );
+}
+
+function Toggle({ opensubmenu, setOpenSubmenu, callback }) {
```
fnwbr, line 51: No need to componentize `ServiceSubmenu.Toggle` at this point I guess. Or do we plan on having one of those menus *without* the toggle?

When you include the toggle within the `ServiceSubmenu` component right away, then it's easy to keep the state (open/closed) in there. Like this:

```js
   const [menuIsOpen
   return (<Header>
        <h1>{ title }</h1>
        <Toggle ...></Toggle>
        { React.Children.map(children, child =>
            React.cloneElement(child, { opensubmenu, setOpenSubmenu }),
        ) }
    </Header>
    );
```

pages/write/[[...roomId]].js
```diff
@@ -84,7 +47,7 @@ export default function Write() {
     const [validatePassword, setValidatePassword] = useState('');
     const [actionSelect, setActionSelect] = useState('');
     const [serviceSpaceId, setServiceSpaceId] = useState();
-    const [openActions, setOpenActions] = useState(false);
+    const [, setOpenActions] = useState(false);
```
Comment (fnwbr) [inline pages/write/[[...roomId]].js:50]: This is also something that's going to always be there when we make use of the action menu. So it should live inside there...

To do this we could pass every "action" that's supposed to be rendered with the corresponding `ServiceSubmenu.Item` like this:

```js
// this replaces the `renderSelectedOption()` method inside of `write/index.js`
const ActionNewAnonymousPad = (<form onSubmit={(e) => { e.preventDefault(); createAnonymousPad(); }}>
                    <input type="text" placeholder={t('pad name')} value={newPadName} onChange={(e) => setNewPadName(e.target.value)} />
                    <button type="submit" disabled={!newPadName}>{ loading ? <LoadingSpinner inverted /> :t('Create pad') }</button>
</form>);

...

<ServiceSubmenu.Item actionComponentToRender={ActionNewAnonymousPad}>{ t('Create new anonymous pad') }</ServiceSubmenu.Item>
```

components/UI/ServiceSubmenu.js
marcel-klasse: I would argue that over the course of redesigning this might be valuable. Potentially desktop view wouldn't use the toggle.

fnwbr: As discussed in person: Desktop vs. mobile does not make a difference for -if- the React component should be a separate component or not. Desktop vs. mobile will be dealt with via CSS.

components/UI/Details.js
```diff
@@ -0,0 +1,23 @@
+import styled from 'styled-components';
+
+const DetailsWrapper = styled.details`
```
fnwbr, line 3: If there's no style in here... why not just write `<details>` in line 15?

components/UI/Form.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+
+const Form = styled.form`
+  /* if first child of parent element (i.e. wrapper), set margin-top */
+  &:not(:first-child) {
+    margin-top: calc(var(--margin) / 1.5);
+  }
+
+  /* for all children except last child, set margin-top */
+  & > * + * {
```
fnwbr, line 10: I'm rambling, feel free to ignore: I don't have a good alternative solution handy right now... but this is basically going back to our old, hard-to-read, hard-to-comprehend and therefore hard-to-maintain style of writing CSS.

It's magically touching everything you put inside of this (allegedly) general-purpose `Form` component...

components/UI/LoadingSpinner.js
```diff
@@ -1,10 +1,10 @@
 import styled from 'styled-components';
 
 const LoadingSpinner = styled.div`
-  display: inline-block;
+  display: block;
   width: 2rem;
   height: 2rem;
-  border-color: ${props => props.inverted ? 'var(--color-bg) transparent var(--color-bg) transparent' : 'var(--color-fg) transparent var(--color-fg) transparent '};
+  border-color: ${props => props.inverted ? 'var(--color-background) transparent var(--color-background) transparent' : 'var(--color-foreground) transparent var(--color-foreground) transparent '};
```
fnwbr, line 7: Generally: This is a good example for a commit that should happen on the `main` branch. The variables have changed there already; it's not really part of this "Refactor navigation" change....

components/UI/ServiceSubmenu.js
```diff
@@ -0,0 +1,93 @@
...
+
+const ToggleButton = styled.button`
+  /* unset globally defined button styles; set height to line-height */
+  width: unset;
+  height: calc(var(--margin) * 1.3);
+  padding: unset;
+  background-color: unset;
+  border: unset;
+
+  /*
```
fnwbr, line 19: Remove please if unused

components/UI/ServiceSubmenu.js
```diff
@@ -0,0 +1,93 @@
...
+
+export function ServiceSubmenu({ title, icon, children }) {
+    const [opensubmenu, setOpenSubmenu] = useState(false); // @TODO stylelint is throwing an error for line 27 if variable is camelCase. disableing stylelint for the line doesn't work for some reason.
+    const handleMenuToggle = () => setOpenSubmenu(!opensubmenu);
+    return (
+        <>
+            <Header>
+                { title && title }
+                <ToggleButton onClick={handleMenuToggle}>
+                    { /* <MenuAdd fill="var(--color-background)" /> */ }
```
fnwbr, line 54: I'll stop commenting further, but please remove any code that's not actually being used...

components/layouts/iframe.js
```diff
@@ -53,6 +54,39 @@ const IframeWrapper = styled.div`
...
+  width: 100%;
+  padding: var(--margin);
+  background: rgb(248 248 248);
+
+  @media (width > 68em) {
+    padding: calc(var(--margin) * 2);
+  }
+
+  /* @TODO: re-factor quick-fix */
+  & h2 {
```
fnwbr, line 69: No need for the `&`

components/layouts/iframe.js
```diff
@@ -53,6 +54,39 @@ const IframeWrapper = styled.div`
...
+  }
+`;
+
+const IframeHeaderButtonWrapper = styled.div`
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: calc(var(--margin));
+  height: 100%;
+
+  & button {
```
fnwbr, line 80: No need for the `&`

components/UI/ServiceTable.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+`;
+
+const TableCell = styled.td`
+  width: 100%;
+  padding: 0;
+  font-size: 85%;
+  text-overflow: ellipsis;
+
+  & a {
```
fnwbr, line 35: No need for the `&`

components/UI/ServiceSubmenu.js
```diff
@@ -0,0 +1,93 @@
...
+  button {
+    padding: calc(var(--margin) / 1.5);
+    text-align: center;
+
+    & + button {
+      margin-top: calc(var(--margin) / 1.5);
+    }
+  }
+
+  /* set margin-bottom on very last child of submenu */
```
fnwbr, line 40: Why not add the margin bottom to the submenu itself?

As in...

```
styled.aside`
    margin-bottom: calc(var(--margin) / 1.5);

    ...
`
```

fnwbr (changes requested):

components/UI/LoadingSpinnerInline.js
```diff
@@ -0,0 +1,11 @@
...
+
+import LoadingSpinner from './LoadingSpinner';
+
+const LoadingSpinnerInline = styled(LoadingSpinner)`
+  display: inline-block;
+  width: calc(var(--margin) * 1.4);
+  height: calc(var(--margin) * 1.4);
+`;
+
+export default LoadingSpinnerInline;
```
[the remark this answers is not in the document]
marcel-klasse: https://github.com/medienhaus/medienhaus-spaces/pull/10/files#diff-f055340701aa99f7442c2bde13bb1ac3c389faa4929cc634e492fb16fe477525R220

components/UI/ServiceSubmenu.js
marcel-klasse: this would push down the table below by default

pages/write/[[...roomId]].js
```diff
@@ -166,132 +141,199 @@ export default function Write() {
...
-        });
-        if (!padId) {
+    const ActionExistingPad = () => {
+        const [padName, setPadName] = useState('');
+        const [padLink, setPadLink] = useState('');
+        const [validLink, setValidLink] = useState('undefined');
+        const [loading, setLoading] = useState(false);
+
+        const handleExistingPad = (e) => {
+            if (getConfig().publicRuntimeConfig.authProviders.write.bypassUrlValidation) {
```
Comment (fnwbr) [inline pages/write/[[...roomId]].js:184]: A bit confused... if `bypassUrlValidation` is true... why do we check if the URL is valid then? Is this supposed to be the other way around? Can you add some commenting and more clear variable names on how this is intended to work?

fnwbr (changes requested):

pages/write/[[...roomId]].js
Comment (fnwbr) [reply pages/write/[[...roomId]].js:184]: I think I understood... maybe a better variable name for the config variable would be something like "`allowThirdPartyPads`" oder "`allowPadsFromOtherSites`"?

Comment (marcel-klasse) [reply pages/write/[[...roomId]].js:184]: we can change the name to make it clearer. but for now (unless we change the validation) literally any link can be saved, though. there's no check if its an actual etherpad link or not.

fnwbr (approved):

PR #11, opened 2022-11-02
Add /sketch service

marcel-klasse:
- Adds implementation of a service page for /sketch (Spacedeck instances)
- Adds authentication provider for said Spacedeck instance
- Automatically syncs previously existing spaces from the underlying Spacedeck server into your Matrix


next.config.example.js
```diff
@@ -16,10 +16,6 @@ module.exports = {
                 baseUrl: 'https://pad.riseup.net/p',
                 api: 'https://pad.riseup.net/mypads/api',
             },
```
fnwbr, line 18: Can you add an example for how to configure /sketch?

marcel-klasse:
**In order for sketch ([spacedeck](https://github.com/spacedeck/spacedeck-open)) to work within an iframe, we need to accept the spacedeck cookie. This creates a problem if the sketch service is not hosted on the same domain as spaces**

pages/sketch/[[...roomId]].js
```diff
@@ -0,0 +1,300 @@
...
+            </IframeLayout.Sidebar>
+            { roomId && content && (
+                <IframeLayout.IframeWrapper>
+                    <IframeLayout.IframeHeader>
+                        <h2>{ matrix.rooms.get(roomId).name }</h2>
+                        <IframeLayout.IframeHeaderButtonWrapper>
+                            <button title={t('Copy pad link to clipboard')} onClick={copyToClipboard}>
+                                <Clipboard fill="var(--color-foreground)" />
+                            </button>
+                            <button title={t('Remove sketch from my library')} onClick={removeLink}>
```
Comment (fnwbr) [inline pages/sketch/[[...roomId]].js:286]: I remember for pads we used "Remove from library" to verbalize that these pads are not actually being "deleted" from the Etherpads server, but only removed from your Matrix account by deleting the Matrix room.
But for Spacedeck sketches we **do actually** delete the sketch, right? Therefore I'd go with "Delete sketch" here instead.

What do you think?

assets/icons/edit.svg
fnwbr, line 1: Let's delete icons that we don't actually use

components/UI/CopyToClipboard.js
```diff
@@ -0,0 +1,26 @@
...
+
+    const copyToClipboard = async () => {
+        navigator.clipboard.writeText(content);
+        setContentCopied(true);
+        await new Promise(r => setTimeout(r, 2000));
+        setContentCopied(false);
+    };
+
+    return (
+        <TextButton title={t('Copy link to clipboard')} onClick={copyToClipboard}>
```
fnwbr, line 19: Make sure that the `title` can also be passed as a parameter in line 7 like this:
```js
const CopyToClipboard = ({ content, title }) => {
```

components/UI/CopyToClipboard.js
fnwbr, line 1: This might have been me but... right now we're not using this component anymore. :')
Can you use this `<CopyToClipboard>` component where it's applicable?

fnwbr (changes requested):
Please check
1. for references to "pads", for example `Copy pad link to clipboard` or `populatePadsfromServer` (easiest is to search on this site https://github.com/medienhaus/medienhaus-spaces/pull/11/files for "pad") and change them to "sketch" accordingly
2. translations to German -- I can still see English words like "Create" when switching to German language

pages/sketch/[[...roomId]].js
```diff
@@ -0,0 +1,300 @@
...
+            callbackDone && callbackDone();
+            setLoading(false);
+
+        };
+
+        return (
+            <Form onSubmit={handleSubmit}>
+                <input type="text" placeholder={t('sketch name')} value={sketchName} onChange={(e) => setSketchName(e.target.value)} />
+                <input type="text" placeholder={t('link to sketch')} value={sketchLink} onChange={handleExistingSketch} />
+                { !validLink && sketchLink !=='' && <span>{ t('Make sure your link includes') }:  { getConfig().publicRuntimeConfig.authProviders.sketch.baseUrl }</span> }
```
Comment (fnwbr) [inline pages/sketch/[[...roomId]].js:245]: This should make use of the `<ErrorMessage` component, like it does for /write

fnwbr (changes requested):

pages/sketch/[[...roomId]].js
Comment (marcel-klasse) [reply pages/sketch/[[...roomId]].js]: yeah that makes sense, maybe we should actually give people the option to do both. So they can remove a sketch from their sketch folder without deleting it (if its a shared sketch for example). 
I _think_ this is addressed in [roomActions](https://github.com/medienhaus/medienhaus-spaces/tree/roomActions ). Or at least the difference between deleting and removing, visually we need to make this clear though as well and maybe stop using the bin icon for removing and only for deleting?

next.config.example.js
```diff
@@ -18,23 +18,42 @@ module.exports = {
                 // api: 'https://your.etherpadserver.org/mypads/api',
                 // spacesGroupName: '/spaces'
             },
+            spacedeck: {
+                path: '/sketch',
+                baseUrl: 'http://spacedeck.localhost',
```
fnwbr, line 23: Can we comment out spacedeck by default?
Similar to how I commented out the `etherpad.api` property -- because it's not a real MyPads API -- I think I'd like to comment this out to make sure that `next.config.example.js` is a _working_ config, if someone would just copy and paste it to `next.config.js` exactly as it is.

Following that logic I guess that `matrixContentStorage` would have to be commented out as well...

What do you think @marcel-klasse?

components/UI/CopyToClipboard.js
```diff
@@ -0,0 +1,26 @@
...
+
+    const copyToClipboard = async () => {
+        navigator.clipboard.writeText(content);
+        setIsContentCopied(true);
+        await new Promise(r => setTimeout(r, 2000));
+        setIsContentCopied(false);
+    };
+
+    return (
+        <TextButton title={title || t('Copy link to clipboard')} onClick={copyToClipboard}>
```
fnwbr, line 19: If I see this correctly then we're missing a translation for "Copy link to clipboard"... that should probably go to `_defaults.json`

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,297 @@
...
+            await createSketchRoom(sketchLink, sketchName);
+            callbackDone && callbackDone();
+            setLoading(false);
+        };
+
+        return (
+            <Form onSubmit={handleSubmit}>
+                <input type="text" placeholder={t('sketch name')} value={sketchName} onChange={(e) => setSketchName(e.target.value)} />
+                <input type="text" placeholder={t('link to sketch')} value={sketchLink} onChange={handleExistingSketch} />
+                { !validLink && sketchLink !=='' && <ErrorMessage>{ t('Make sure your link includes') }:  { getConfig().publicRuntimeConfig.authProviders.spacedeck.baseUrl }</ErrorMessage> }
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:244]: Can we reuse the `Make sure your link includes "{{url}}"` error message from the /write PR?
And therefore move the translation of that to the `_defaults.json` file?

public/locales/de/sketch.json
```diff
@@ -0,0 +1,9 @@
+{
+    "Make sure your link includes": "Der Link muss folgende Adresse beinhalten",
+    "Add existing sketch": "Bestehenden Sketch hinzufügen",
+    "Create Sketch": "Sketch erstellen",
+    "link to sketch": "Link zum sketch",
```
fnwbr, line 5: Deutsch: Konsistenz -- Schreiben wir "Sketch" groß oder "sketch" klein?

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,297 @@
...
+            // we check if the link is valid for the service (has the same base url)
+            if (e.target.value.includes(getConfig().publicRuntimeConfig.authProviders.spacedeck.baseUrl)) setValidLink(true);
+            else setValidLink(false);
+            setSketchLink(e.target.value);
+        };
+
+        const handleSubmit = async (e) => {
+            setLoading(true);
+            e.preventDefault();
+            await createSketchRoom(sketchLink, sketchName);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:235]: After this line: Shouldn't we `router.push(...` the user to the added sketch after this was submitted?
We do it for other /sketch boards (see `ActionNewSketch`) and also when adding existing /write pads...

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,297 @@
...
+            callbackDone && callbackDone();
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:236]: And looking at `handleExistingPadSubmit` from the /write PR once again:
I think we want to `setSketchLink('')` here after this line.

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,297 @@
...
+
+    const removeLink = async () => {
+        setRemovingLink(true);
+        const remove = await spacedeck.deleteSpaceById(content.body.substring(content.body.lastIndexOf('/') + 1)).catch((e) => console.log(e));
+        if (!remove) {
+            setRemovingLink(false);
+
+            return;
+        }
+        if (!remove.ok) {
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:183]: So this `if()` and the `if()` above end up doing the exact same thing... couldn't we write something like this then?
```js
if (!remove || !remove.ok) {
    ...
}
```

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,297 @@
...
+        // @TODO callback function to give user feedback when removing on the server fails
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:184]: Regarding this `@TODO`: I know some people are not the biggest fan of `alert()`, **but** in lack of better solutions how about telling the user that deleting the link failed and letting them know that the /sketch deck -still- exists?

I guess this is an edge case and in most cases deleting will work just fine anyways.

fnwbr (changes requested):

pages/spacedeck/SketchLinkEntry.js
fnwbr, line 1: Can we rename this component according to the new guidelines? This should be a `SpacedeckLinkEntry` I guess. Also:

Please compare this component to how I refactored `EtherpadListEntry.js` and make it a "display only" component. By that, I mean, that this component should not be containing any "logic" code on how to... get the name of the sketch (or Matrix room), but it should only receive it as a parameter.

The loading of all data should happen in `spacedeck/[[...roomId]].js` -- and then all necessary components rendered should receive the data from there.

❇️ Bonus for later (not asking you to do this as part of this Pull Request): Once both, `EtherpadListEntry.js` and `SpacedeckListEntry.js` are going to be -almost- the same component in the end, we might ✨ generalize ✨ them into something like a general... `ServiceLink.js` component?! 😉

marcel-klasse: went ahead and did the bonus since it wasn't really any more effort 👍

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+
+                }
+            </IframeLayout.Sidebar>
+            { roomId && content && (
+                <IframeLayout.IframeWrapper>
+                    <IframeLayout.IframeHeader>
+                        <h2>{ matrix.rooms.get(roomId).name }</h2>
+                        <IframeLayout.IframeHeaderButtonWrapper>
+                            <CopyToClipboard title={t('Copy sketch link to clipboard')} content={content.body} />
+                            <button title={t('Delete sketch from my library')} onClick={removeLink}>
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:295]: Missing DE translation for "Delete sketch from my library"

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+                                { removingLink ? <LoadingSpinnerInline /> : <Bin fill="var(--color-foreground)" /> }
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:296]: NAB (not a blocker): I like `isDeletingPad` like in **/write** as a variable name better... `removingLink` is making me think like... what link?!? And the `is` prefix is making it more clear that it's a boolean indicating if we're currently in the process of doing so.

public/locales/de/_defaults.json
```diff
@@ -2,5 +2,9 @@
     "Password": "Passwort",
     "Select action": "Aktion wählen",
     "What would you like to do?": "Was möchtest Du tun?",
-    "{{error}}, please wait {{time}} seconds.": "{{error}}, bitte warte {{time}} sekunden."
+    "{{error}}, please wait {{time}} seconds.": "{{error}}, bitte warte {{time}} sekunden.",
```
fnwbr, line 5: ... "Sekunden" with a capital S, please ...

components/UI/ServiceLink.js
```diff
@@ -4,19 +4,21 @@ import Link from 'next/link';
...
+const ServiceLink = forwardRef(({ roomId, name, selected, path, passwordProtected }, ref) => {
+    const { t } = useTranslation();
 
     return (
         <ServiceTable.Row>
-            <ServiceTable.Cell selected={selected}><Link ref={ref} disabled href={`/write/${roomId}`}>{ padName }</Link></ServiceTable.Cell>
+            <ServiceTable.Cell selected={selected}>
+                <Link ref={ref} disabled href={`/${path}/${roomId}`}>{ name }</Link>
+            </ServiceTable.Cell>
             { /* Show a lock icon if this is a password protected pad */ }
```
fnwbr, line 22: The comment "password protected pad" is wrong now that it's a generalized component

components/UI/CopyToClipboard.js
```diff
@@ -0,0 +1,26 @@
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import Clipboard from '../../assets/icons/clipboard.svg';
+import TextButton from './TextButton';
+
+const CopyToClipboard = ({ content, title }) => {
+    const [isContentCopied, setIsContentCopied] = useState(false);
```
fnwbr, line 8: NAB: `wasContentCopied` is a more correct & declarative variable name, since there's no asynchronous action happening here. Instead the content was copied to the clipboard immediately after invoking the function.
We just set a timer ourselves... the `is` prefix IMHO makes sense when performing asynchronous actions like making a request to a server and waiting for the response.

components/UI/ServiceLink.js
```diff
@@ -4,19 +4,21 @@ import Link from 'next/link';
...
-const EtherpadListEntry = forwardRef(({ roomId, padName, passwordProtected, selected }, ref) => {
-    const { t } = useTranslation('write');
+const ServiceLink = forwardRef(({ roomId, name, selected, path, passwordProtected }, ref) => {
+    const { t } = useTranslation();
 
     return (
         <ServiceTable.Row>
-            <ServiceTable.Cell selected={selected}><Link ref={ref} disabled href={`/write/${roomId}`}>{ padName }</Link></ServiceTable.Cell>
+            <ServiceTable.Cell selected={selected}>
+                <Link ref={ref} disabled href={`/${path}/${roomId}`}>{ name }</Link>
```
fnwbr, line 20: Let's generalize this a bit more and only pass `path` (and **not** `roomId` anymore) as a parameter to `ServiceLink`.
`ServiceLink` should not automatically assume that it's always `/<x>/<roomId>` -- instead the consuming page or component will take care of that logic itself by passing `path={'/etherpad/${roomId}'}` as a parameter instead.

pages/etherpad/[[...roomId]].js
```diff
@@ -6,16 +6,16 @@ import { useRouter } from 'next/router';
 
 import { useAuth } from '../../lib/Auth';
 import { useMatrix } from '../../lib/Matrix';
-import EtherpadListEntry from './EtherpadListEntry';
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:9]: Now that this file is not being used anymore... can we delete `EtherpadListEntry.js`?

pages/etherpad/[[...roomId]].js
```diff
@@ -213,10 +211,11 @@ export default function Etherpad() {
                         { getConfig().publicRuntimeConfig.authProviders.etherpad.myPads?.api && !serverPads && <ErrorMessage>{ t('Can\'t connect to the provided /write server. Please try again later.') }</ErrorMessage> }
                         <ServiceTable>
                             { matrix.spaces.get(matrix.serviceSpaces.etherpad).children?.map(writeRoomId => {
-                                return <EtherpadListEntry
+                                return <ServiceLink
                                     key={writeRoomId}
                                     roomId={writeRoomId}
-                                    padName={_.get(matrix.rooms.get(writeRoomId), 'name')}
+                                    name={_.get(matrix.rooms.get(writeRoomId), 'name')}
+                                    path={getConfig().publicRuntimeConfig.authProviders.etherpad.path?.replace(/[<>\s/:]/g, '') || 'etherpad'}  // sanitizing the string just in case of any forbidden url characters
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:218]: This shouldn't be necessary... one should sanitize user input, yes. But sanitizing a value that was put in by a sysadmin/developer/engineer shouldn't be necessary. Also, let's take care of this in a separate PR please. It's basically the issue that I opened up here, that you're starting to address: https://github.com/medienhaus/medienhaus-spaces/issues/30

And if I understood the NextJS documentation correctly, it should be totally fine to just put `etherpad` here. The router will redirect you to, let's say `write` for example, if that's the path alias that was set up in the config file.

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+        selectedPadRef.current?.focus();
+    }, [roomId]);
+
+    useEffect(() => {
+        let cancelled = false;
+
+        const startLookingForFolders = async () => {
+            if (matrix.initialSyncDone) {
+                try {
+                    setServiceSpaceId(matrix.serviceSpaces.spacedeck);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:53]: Correct me if I am wrong, but I don't think we need to store this service space ID in a separate state.
Check out what I did in https://github.com/medienhaus/medienhaus-spaces/commit/7258c4e409894c3d7cce879d0c6dc5d64dd02e3a -- I think we can do the same here... Let me know if it doesn't work or if you have any questions.

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+            const updateStructure = async (object, parent) => {
+                for (const sketch of Object.values(object)) {
+                    // if the element does not have an id key, we can safely assume it is a key of a folder from recursion and skip it.
+                    if (!sketch.id) continue;
+                    // now we can check if a sketch already exists in matrix and if so we can skip this sketch.
+                    if (matrixSketches[sketch.id]) {
+                        // we check if the names of our sketches are still matching on the matrix server and on the sketch server
+                        if (sketch.name !== matrixSketches[sketch.id].name) {
+                            // eslint-disable-next-line no-undef
+                            if (process.env.NODE_ENV === 'development') console.log('changing name for ' + matrixSketches[sketch.id]);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:107]: Use `logger.debug` (or `logger.info`) as per https://github.com/medienhaus/medienhaus-spaces/pull/36

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+        let cancelled = false;
+
+        !cancelled && setContent(matrix.roomContents.get(roomId));
+
+        return () => cancelled = true;
+    }, [matrix.roomContents, roomId]);
+
+    async function createSketchRoom(link, name, parent = serviceSpaceId) {
+        // eslint-disable-next-line no-undef
+        if (process.env.NODE_ENV === 'development') console.debug('creating room for ' + name);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:166]: Same as above: use `logger.debug()` instead

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+    const [content, setContent] = useState(matrix.roomContents.get(roomId));
+    const [syncingServerSketches, setSyncingServerSketches] = useState(false);
+    const [isSpacedeckServerDown, setIsSpacedeckServerDown] = useState(false);
+    const path = getConfig().publicRuntimeConfig.authProviders.spacedeck.path?.replace(/[<>\s/:]/g, '') || 'spacedeck';
+
+    const spacedeck = auth.getAuthenticationProvider('spacedeck');
+
+    // Whenever the roomId changes (e.g. after a new sketch was created), automatically focus that element.
+    // This makes the sidebar scroll to the element if it is outside of the current viewport.
+    const selectedPadRef = useRef(null);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:42]: Name this variable `selectedSketchRef` instead

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,309 @@
...
+
+        return () => {
+            cancelled = true;
+        };
+    }, [spacedeck]);
+
+    useEffect(() => {
+        let cancelled = false;
+
+        !cancelled && setContent(matrix.roomContents.get(roomId));
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:159]: I _think_ we can stop doing this as well... check the first commit of https://github.com/medienhaus/medienhaus-spaces/pull/45 and see if it works here as well...

fnwbr (changes requested):

public/locales/de/spacedeck.json
```diff
@@ -0,0 +1,11 @@
+{
+    "Make sure your link includes": "Der Link muss folgende Adresse beinhalten",
+    "Add existing sketch": "Bestehenden Sketch hinzufügen",
+    "Create Sketch": "Sketch erstellen",
```
fnwbr, line 4: What's the difference from this to line 6?

public/locales/de/spacedeck.json
```diff
@@ -0,0 +1,11 @@
+{
+    "Make sure your link includes": "Der Link muss folgende Adresse beinhalten",
```
fnwbr, line 2: This line is also not being used anymore, or is it?

public/locales/de/spacedeck.json
```diff
@@ -0,0 +1,11 @@
...
+    "Add existing sketch": "Bestehenden Sketch hinzufügen",
+    "Create Sketch": "Sketch erstellen",
+    "link to sketch": "Link zum Sketch",
+    "Create sketch": "Sketch erstellen",
+    "Remove sketch": "Sketch löschen",
```
fnwbr, line 7: `Remove sketch` is not being used anywhere

pages/spacedeck/actions/AddExistingSketch.js
```diff
@@ -0,0 +1,55 @@
...
+            }));
+        router.push(`/${path}/${roomId}`);
+        setSketchLink('');
+        callbackDone && callbackDone();
+        setLoading(false);
+    };
+
+    return (
+        <Form onSubmit={handleSubmit}>
+            <input type="text" placeholder={t('sketch name')} value={sketchName} onChange={(e) => setSketchName(e.target.value)} />
```
fnwbr, line 45: I think we're missing a DE translation for this, as in German we're always capitalizing the word "Sketch" and in English we're using the lowercase variant. When looking at this UI in German this input would be the only one saying `sketch name`, when all other inputs and buttons refer to `Sketch` (with a capital S).

pages/spacedeck/[[...roomId]].js
```diff
@@ -0,0 +1,226 @@
...
+import IframeLayout from '../../components/layouts/iframe';
+import { ServiceTable } from '../../components/UI/ServiceTable';
+import CopyToClipboard from '../../components/UI/CopyToClipboard';
+import ServiceLink from '../../components/UI/ServiceLink';
+import CreateNewSketch from './actions/CreateNewSketch';
+import AddExistingSketch from './actions/AddExistingSketch';
+
+export default function Spacedeck() {
+    const auth = useAuth();
+    const matrix = useMatrix(auth.getAuthenticationProvider('matrix'));
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:25]: `useMatrix()`

fnwbr (changes requested):

PR #23, opened 2023-07-13
Change naming of `etherpadMyPads` to `etherpad` and add customisable path name

marcel-klasse:
This PR enables users to define a custom name for their etherpad route, which will be displayed both in the adress bar, as well as in the navigation.  

 A custom name can be defined in the `next.config.js `under `etherpad.path`.

`rewrites()` within `next.config.js` makes sure the correct name is displayed on the users end. Internally we can use the application name, in this case etherpad, to reference the correct application. 

This arguably makes development easier and gives users the freedom to choose the name of their route.

The extension of `...MyPads` is dropped since it is only applicable if `etherpad.api` is supplied. (this could also be renamed to `etherpad.myPads`)

The application folder within matrix is called `etherpad` regardless of the supplied path name.


fnwbr:
> The extension of `...MyPads` is dropped since it is only applicable if `etherpad.api` is supplied. (this could also be renamed to `etherpad.myPads`)

Don't like this change for the reason that I tried to clarify, that you **need** the MyPads extension, in order for this authentication provider to work. Literally _every_ API call in the `lib/auth/Etherpad(MyPads)AuthProvider.js` file works against the MyPads API, and _not_ against that claimed "Etherpad" instance -- I find that to be misleading.

> (this could also be renamed to `etherpad.myPads`)

Yeah, I think that's a good idea. Move MyPads-specific configuration values into another object, like...

```js
etherpad: {
    path: '/write',
    baseUrl: 'https://pad.riseup.net/p',
    myPads: {
        api: 'https://your.etherpadserver.org/mypads/api',
        spacesGroupName: '/spaces',
    },
},
...
```

lib/Auth.js
```diff
@@ -87,6 +87,8 @@ function useAuthProvider() {
         window.localStorage.removeItem('mx_hs_url');
         window.localStorage.removeItem('mx_user_id');
         window.localStorage.removeItem('mx_device_id');
+        window.localStorage.clear()
```
fnwbr, line 90: This does not make sense. Why manually remove certain items in the lines above, when you clear the whole storage afterwards?

I also prefer to manually remove only those entries, which I also purposefully set beforehand. If there's something in the local storage that shouldn't be there, then let's try to find a place in the code that makes more sense to remove it from there.

fnwbr (commented):
All in all I think this makes the codebase more obscure: 61 lines of code changed, only to allow typing "/write" just like one would have done beforehand.

People who are familiar with Next.js will
1. click around the app
2. see something in `/write/<blablabla>` they want to change
3. go to the `pages` directory in the source code
4. look for the "write" file (or directory)
5. 😢 and not find one -- because what they're looking for is inside the "etherpad" directory.

> This arguably makes development easier and gives users the freedom to choose the name of their route.

I agree with the latter, but I don't agree with the former. It's doing the opposite.

lib/Auth.js
marcel-klasse: I agree, either just `clear()` or individual items. Personally I prefer just `clear()` since it's only one line of code but it's not a strong preference.

I will fix this in a future commit and go back to individual items.

marcel-klasse:
> Don't like this change for the reason that I tried to clarify, that you need the MyPads extension, in order for this authentication provider to work. Literally every API call in the lib/auth/Etherpad(MyPads)AuthProvider.js file works against the MyPads API, and not against that claimed "Etherpad" instance -- I find that to be misleading.

Hm I guess you're right when it comes to the auth provider. In that case I would actually suggest callling it simply `MyPadsAuthProvider`, since that is essentially what it is doing, although `EtherpadMyPadsAuthProvider` is clearer, it is also at some point quite hard to read. Or maybe `Ep_MyPadsAuthProvider` like the [repo](https://framagit.org/framasoft/Etherpad/ep_mypads)
`/pages/etherpad.js` on the other hand could remain like this. since you can use it without MyPads. 

> Yeah, I think that's a good idea. Move MyPads-specific configuration values into another object, like...

👍 Let's implement this regardless of which naming scheme we will be using in the end.

> All in all I think this makes the codebase more obscure: 61 lines of code changed, only to allow typing "/write" just like one would have done beforehand.
> 
> People who are familiar with Next.js will
> 
>     click around the app
>     see something in /write/<blablabla> they want to change
>     go to the pages directory in the source code
>     look for the "write" file (or directory)
>     😢 and not find one -- because what they're looking for is inside the "etherpad" directory.
> 

I can see your point, but I believe that if we follow this logic we should go back to the initial naming scheme of `writeAuthProvider` and not mix it with the actual repository names.  

Actually, maybe by default the path should be called etherpad (if no other path name like '/write' is given). If people want a different name for their route they can esaily configure it, instead of us forcing an arbitraitily chosen name on them. 

It would also make it easier with future plugins, say if there was a plugin for [outline](https://github.com/outline/outline)  and etherpad, `/write` coud be a little misleading since this could mean both outile and etherpad. 


Obviously none of this is set in stone, but my suggestion would be to either hardcode the descriptive terms like `write` or make them completely optional and customisable.

fnwbr:
> Hm I guess you're right when it comes to the auth provider. In that case I would actually suggest callling it simply `MyPadsAuthProvider`, since that is essentially what it is doing, although `EtherpadMyPadsAuthProvider` is clearer, it is also at some point quite hard to read. Or maybe `Ep_MyPadsAuthProvider` like the [repo](https://framagit.org/framasoft/Etherpad/ep_mypads)
`/pages/etherpad.js` on the other hand could remain like this. since you can use it without MyPads. 

👍 

> Actually, maybe by default the path should be called etherpad (if no other path name like '/write' is given). If people want a different name for their route they can esaily configure it, instead of us forcing an arbitraitily chosen name on them. 
> 
> It would also make it easier with future plugins, say if there was a plugin for [outline](https://github.com/outline/outline)  and etherpad, `/write` coud be a little misleading since this could mean both outile and etherpad. 

👍 - I can get down to this. Stick to technical terms all the way, and let the user add redirects or custom translations if they want to change their "product reference name".

This might make translation a bit trickier though... in many places we still refer to `/write` or we do `router.push('/write')`... have you tested your latest changes with that?

> Obviously none of this is set in stone, but my suggestion would be to either hardcode the descriptive terms like `write` or make them completely optional and customisable. 

I still would have been fine with
```
- pages
    - write
    - sketch
    - ...
```
... on one hand, and ...
```
- lib
    - auth
        - MyPadsAuthProvider
        - SpacedeckAuthProvider
        - ...
```
on the other hand. One referring to the more product-y name, and the other side referring to the very technical term.
Our "product /write", the medienhaus-spaces implementation, relies on the technical basis of MyPads.
Our "product /sketch" is our implementation of what Spacedeck offers under the hood, or via their API.

But as I said above, personally I can agree to sticking to technical terms -everywhere- as well. I had assumed that you'd dislike that more.

PR #25, opened 2023-07-14
Add explore actions and dedicated /actions route

andirueckel:
Not sure if I got it right, but this should make all actions from `roomActions` branch available via the new `/actions` route; the corresponding files were moved from `pages/explore/*` to `pages/actions/*` to not conflict with the current implementation and feature set of the `/explore` route.

As these actions will probably be implemented in a context-sensitive way, maybe via context menus on individual elements, and as these actions might probably be useful on other routes than `/explore`, I figured it more adequate, and less prone to merge-conflicts, to expose the current implementation via a new `/actions` route for now.

The changes were _not really_ — due to missing room memberships or room permissions of my dev user account — tested with the following `next.config.js` configuration:

```
const WebpackConfig = require('./webpack.config.js');

// eslint-disable-next-line no-undef
module.exports = {
    publicRuntimeConfig: {
        name: '/spaces',
        authProviders: {
            matrix: {
                baseUrl: 'https://dev.medienhaus.udk-berlin.de',
                api: 'https://api.dev.medienhaus.udk-berlin.de',
                allowCustomHomeserver: true,
            },
            matrixContentStorage: {
                baseUrl: 'https://dev.medienhaus.udk-berlin.de',
            },
            etherpad: {
                path: '/write',
                baseUrl: '//etherpad.localhost/p',
            },
        },
        contextRootSpaceRoomId: '!gzsKJXOMipzIxsoqYk:dev.medienhaus.udk-berlin.de',
        templates: {
            context: '!DKzCQzwwAaeXIVkLRh:dev.medienhaus.udk-berlin.de',
        },
        account: {
            allowAddingNewEmails: true,
        },
        chat: {
            pathToElement: '//localhost/element',
        },
    },
    async rewrites() {
        return [
            {
                source: this.publicRuntimeConfig.authProviders.etherpad.path,
                destination: '/etherpad',
            },
            {
                source: this.publicRuntimeConfig.authProviders.etherpad.path + '/:roomId',
                destination: '/etherpad/:roomId',
            },
        ];
    },
    webpack: WebpackConfig,
};
```


robertschnuell:
dont know if it makes sense to merge that already together. would say lets work in this PR until we really got those two branches combined. For now the 'old' expore, which was just a placeholder page with the `ContextMultiLevelSelect` Component in place got renamed to `actions` as a `page`. 
I would see the `actions` functionallity more like a global component and therefor would put it into `components`. 
As the core functionallity of the `actions` component should be an additional interface to interact with:
1. matrix id's
2. user account id's

maybe we have to split this even up into two components, even thou the inteface could be partly the same just the offered functions inside this overlay would be different. 

so let me know if I should work in this PR to refactor the `actions` functions into `components` and trying to apply changes to them so that they will work with the new `explore` interface and not only with the `ContextMultiLevelSelect`, or if I should create a new branch for that, which we will mergen then later on in the `expore` branch. 
I just thought that the current `expore` branch is more or less clean'ish and bringing now a temporarily `actions` route in there to just work directly in this `expore` branch would create a lot of new chaos. But thats only my 5 cents, open for all kind of solutions.

robertschnuell:
Previousely suggested changes were merged into this current PR. feel free to revert. At some point the `actions` need to be relocated once more or renamed into `roomIdHandle` and `userIdHandle` to get a propper naming.
the `actions` functionallity should be therefore one levle above, where `roomIdHandle` or `userIdHandle` could be a part of it, but also all application specific second actions should be part of if.

PR #27, opened 2023-07-20
layout improvements

andirueckel:
We’ll use a default `font-size` of `16px` as base font size; not going smaller than 16 px prevents iOS Safari from [zooming](https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/) on form elements like `input`; we won’t be dynamically scaling the font size anymore based around `calc()` and `vmin`.

For the navigation menu _open_ and _close_ button, we’ll use additional SVG icons; this was especially necessary for the narrow width layout, when the navigation menu _open_ and e.g. `/write` service table submenu buttons with their respective icons are positioned directly below each other.

The media queries for current width-dependent layout was reduced from three to two, eliminating the `laptopAndAbove` query for widths starting from `1280px` that was only used to change spacing, which we won’t be doing any longer.

The hardcoded width for the navigation sidebar and service table sidebar was increased and based on the `ch` unit, i.e. [“[the] advance measure (width) of the glyph "0" of the element's font.”](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units)

And the language selector was changed in style and position to better integrate with the navigation menu.

**Note:** Navigation on narrow screen needs re-factoring and re-styling; we’ll do this later™ once `/explore` is finished. And maybe we should introduce an intermediate layout that hides the navigation sidebar but not yet the service table sidebar.

Screenshot 2023-07-20 at 12 24 51


andirueckel:
**Note:** Navigation on narrow screen needs re-factoring and re-styling; we’ll do this later™ once `/explore` is finished.

fnwbr:
Me gusta ❤️ 

Only thing I have to mention: the burger menu icon is off by a few pixels now:
image

I know you see what I mean, but: I think the 3 horizontal lines should start one or two pixels more to the top, to be perfectly in line with the base line of the page title / app name.

andirueckel:
@fnwbr Very true. We should, at some point, visually balance the alignment by re-positioning the icon(s) and/or respective button(s); _down the branches_, the `ToggleButton` _styled-component_ is defined and used in three different files to toggle the 1. _navigation menu_, 2. _service table sub-menu_, and 3. _context actions_; hence we could create a proper `ToggleButton` component with respective props being used whenever in need for a toggle button.

Furthermore, we have other — visually identical — action buttons which don’t toggle something, but uni-directionally *do* something; for example the *copy URL* or *delete* buttons in the the `IframeHeader`; in case of `/explore`, two action buttons are positioned directly next to the _context action_ toggle button. Therefore, it might be reasonable to either introduce another button component for these action buttons, or only create *one* button component with respective functionality and props to be used as either action or toggle button.

PR #31, opened 2023-07-26
Dark mode

fnwbr:
Brings back dark mode and also some smaller visual... REFINEMENTS 🧑‍🍳


fnwbr:
This still needs refinement and some work. But it basically works:
image

andirueckel:
Current state of the dark mode:

Screenshot 2023-11-16 at 11 14 43

andirueckel:
Current state of the dark mode:

Screenshot 2023-11-16 at 14 50 56

Current state of the light mode:

Screenshot 2023-11-16 at 15 00 23

NOTE: The colors in the animated gifs are somewhat pale, but you get the idea …

We probably want to display the notification indicator for `/chat` not only for direct messages but also for messages in rooms and maybe even invites (which are already shown on `/dashboard`)? @marcel-klasse Would you mind extending the [respective condition](https://github.com/medienhaus/medienhaus-spaces/pull/31/commits/c094d3c3c529bf42128e0035328da8c685e01066#diff-a3e00a65d0d2f80dcd0a94179a4c3ac1d2ffdba93d1f5f73f19d28fe8d78f7a1R61)?

fnwbr:
> We probably want to display the notification indicator for /chat not only for direct messages but also for messages in rooms and maybe even invites (which are already shown on /dashboard)? @marcel-klasse Would you mind extending the [respective condition](https://github.com/medienhaus/medienhaus-spaces/pull/31/commits/c094d3c3c529bf42128e0035328da8c685e01066#diff-a3e00a65d0d2f80dcd0a94179a4c3ac1d2ffdba93d1f5f73f19d28fe8d78f7a1R61)?

Yes, that's a bit tricky.

We basically want the counter to count pending
- direct messages
- chat room messages
- invites for the two above (but not invites for other things, such as contexts or pads or whatever...)

We do have a similar filtering logic in place already: https://github.com/medienhaus/medienhaus-spaces/blob/4108567e750d6bbd920bfa508407274276697cfc/pages/chat/%5B%5B...roomId%5D%5D.js#L145-L158

We might just need to modify that to fit our needs above. That way we'll end up with a **proper notification indicator for /chat**.

--

Given that has nothing to do with the dark mode though: what's next? Would you feel comfortable merging already?

marcel-klasse (approved):

PR #35, opened 2023-07-27
Use the element home screen to create new chats

marcel-klasse:
Display elements home screen by default when no chat is selected as a first and simple solution to create chats.


fnwbr (changes requested):
Breaks /chat on mobile: I always land on this screen after clicking on **/chat** instead of seeing my list of contacts or rooms:
image

PR #38, opened 2023-07-27
Use element home screen to create new chats by default

aofn:
Displays elements home screen by default on desktop browsers and allows toggling on mobile view.
Allowing users to create DM's and group chats.


fnwbr (commented):
Quick question before going into details of the code changes: Can we not make this happen with CSS media queries _only_?

Detecting the device type with JS is something we never did in any previous React applicatio  (and I have a feeling that we never should do),  when we have the option to change things with CSS only.

pages/chat/[[...roomId]].js
```diff
@@ -169,15 +225,25 @@ export default function RoomId() {
...
                 <br />
             </IframeLayout.Sidebar>
-            { roomId && (
-                <IframeLayout.IframeWrapper>
-                    <iframe src={`${getConfig().publicRuntimeConfig.chat.pathToElement}/#/room/${roomId}`} ref={iframe} />
-                </IframeLayout.IframeWrapper>
-            ) }
+
+            { (deviceType !== 'mobile' || (deviceType === 'mobile' && roomId) || (deviceType === 'mobile' && !isRoomListVisible)) && <IframeLayout.IframeWrapper>
+                <IframeLayout.IframeHeader>
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:236]: I think we don't want the `<IframeLayout.IframeHeader>` component in /chat anymore... it looks like this now:
Screenshot 2023-08-16 at 15-11-42 _spaces

... with two headers. When one is just enough.

Maybe check out what /chat looks like on `main` once more...

fnwbr (changes requested):

aofn:
image
not sure why this is happening. this is also already happening in main

lib/auth/MatrixAuthProvider.js
```diff
@@ -112,21 +112,22 @@ class MatrixAuthProvider {
      * @param {string} type - The type of the room or space.
      * @param {string} template - The template for the room or space.
      * @param {string} parentId - The ID of the parent space, if any.
+     * @TODO: Missing some @param declarations
      * @returns {Promise<Object>} A promise that resolves to the created room or space.
      * @throws {Error} If there was an error creating the room or space.
      */
-    async createRoom(name, isSpace, topic, joinRule, type, template, parentId) {
+    async createRoom(name, isSpace, topic, joinRule, type, template, parentId, application, visibility, historyVisible, encrypted) {
```
fnwbr, line 119: Where are we making use of any of these new parameters?
If we don't use them, let's not add them.

I guess this is a leftover from some previous concepts?

lib/auth/MatrixAuthProvider.js
```diff
@@ -169,7 +180,8 @@ class MatrixAuthProvider {
         const medienhausMetaEvent = {
             type: type,
             template: template,
-            version: '0.4',
+            application: application,
```
fnwbr, line 183: What's the difference between `template` and `application`?

At the same time, like I said above: I think we don't actually make use of this new "application" parameter anywhere. So let's remove it again?

pages/chat/[[...roomId]].js
```diff
@@ -169,30 +186,112 @@ export default function Chat() {
...
                 </details>
                 <br />
                 <details open>
-                    <summary><h3 style={{ display: 'inline-block', marginBottom: '1rem' }}>{ t('Rooms') }</h3></summary>
-                    { otherRooms && otherRooms.map((room) => <SidebarListEntry key={room.roomId} room={room} />) }
+                    <Summary><h3>{ t('Rooms') }</h3></Summary>
+                    <ServiceTable>
+                        <ServiceTable.Body>
+                            { otherRooms && otherRooms.map((room) => {
+                                return <ServiceTable key={room.roomId}><SidebarListEntry key={room.roomId} room={room} selected={room.roomId === roomId} /></ServiceTable>;
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:254]: 🤔 

A `<ServiceTable>` inside of a `<ServiceTable>` (line 251)?
This results in this double-border again:

image

fnwbr (changes requested):
@aofn There are a few things that I can use your help on. I've left comments on the diff.

pages/chat/[[...roomId]].js
```diff
@@ -136,16 +115,54 @@ export default function Chat() {
...
                     .mx_RoomView_timeline_rr_enabled .mx_EventTile[data-layout=group] .mx_ThreadSummary_icon { margin-right: unset; }
+
+                    /* modal overlay when creating rooms */
+
+                    .mx_Dialog { width: 80% }
+                    .mx_CreateRoomDialog.mx_Dialog_fixedWidth { width: 100% }
+                    .mx_Dialog_fixedWidth { width: 100% }
+                    .mx_Dropdown_menu { max-width: 100% }
+
+                    #mx_JoinRuleDropdown__public { display: none }
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:165]: Let's add comments to each of these lines on what they intend to do.
This is especially helpful for later; we do not control if/how/when Element changes their markup. So we might need to make adjustments to all of these CSS rules whenever we update Element. Therefore it's important to remember everything these lines intend to do, so one can later double-check if they still do what they are supposed to do.

pages/chat/[[...roomId]].js
```diff
@@ -136,16 +115,54 @@ export default function Chat() {
...
                     /* Make the "right panel" cover the full screen */
-                    .mx_RightPanel { position: fixed; left: 0; right: 0; bottom: 0; top: 0; z-index: 999999; }
+                    .mx_RightPanel { position: fixed; left: 0; right: 0; bottom: 0; top: 0; z-index: 999999; 
+                    }
+
+                @media ${breakpoints.phoneOnly} {
+                    .mx_RoomHeader { padding: 1rem var(--RoomView_MessageList-padding) }
 
-                    /* More breathing room in the main timeline of a chat */
-                    .mx_RoomView_timeline_rr_enabled .mx_EventTile[data-layout=group] .mx_EventTile_line,
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:146]: Why did you remove these?

pages/chat/[[...roomId]].js
Comment (fnwbr) [reply pages/chat/[[...roomId]].js:165]: Also, if the intention of this specific line is to disallow creating public rooms: Why is it inside a `@media ${breakpoints.phoneOnly} {`? Do we only want to disallow creating public rooms on smartphones? If so, why?

fnwbr (changes requested):

pages/chat/[[...roomId]].js
Comment (fnwbr) [reply pages/chat/[[...roomId]].js:165]: Was the intention to take away this option from users for now?
image

fnwbr:
Alright, I've basically re-implemented the same feature within a few lines of code over in https://github.com/medienhaus/medienhaus-spaces/pull/38.
Feel free to close this PR, **or** modify it to still bring over the changes you did for `<ServiceLink>` -- which I think we want to have!

But maybe it might be easier to close this PR and start from `main` again, and porting the changes manually by just copy-pasting from the diff of this PR... up to you.

pages/chat/[[...roomId]].js
```diff
@@ -169,30 +186,112 @@ export default function Chat() {
...
+                                </Icon>
+                            </MobileBackButton>
+                        </DefaultLayout.IframeHeaderButtonWrapper>
+                    </DefaultLayout.IframeHeader> }
+
+                { /* The iframe can take a long time to load which is why we show a loading spinner.
+                The style hack is needed because setting 'display' to 'none' creates an error.
+                Element performs a few compatability checks before onLoaded fires.
+                Therefore, Element throws an error, saying the current browser is not supported.
+                Setting the height to 0px centers the loading spinner before the iframe is ready.
```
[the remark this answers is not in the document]
Comment (aofn) [reply pages/chat/[[...roomId]].js:285]: hm not sure why the loading spinner didn't show up for you. 
Generally I think we really need to show a loading spinner before the iframe is ready. It just feels like a bug sometimes, otherwise.
I've opened an issue for this in the meantime #119

fnwbr:
Closing this; all handled... what I commented on in my last remark was done in https://github.com/medienhaus/medienhaus-spaces/commit/921d60f4002cf460bf05f564de0d9800bf50b60a

PR #40, opened 2023-08-17
Move service submenu actions into their own directory

fnwbr:
I think this is going to help keeping an overview with the repository, and reduces that enormous `[[...roomId]].js` file to have fewer lines... breaks it up a little.

Aside from splitting up these components into their own files this also helps rendering performance: We should defer from creating sub-components inside components. So we should avoid this:

```js
const MotherComponent = () => {
    ...

    const ChildOne = () => {
        return <div>Child1</div>;
    }
    const ChildTwo = () => {
        return <div>Child2</div>;
    }

    return (
        <ChildOne />
        <ChildTwo />
    );
}
```

Every time that `MotherComponent` gets re-rendered, it will also create those child components from scratch.
Even though they don't actually change. Like in this case with the `/etherpad` page: the service submenu action forms (e.g. to create a new anonymous pad, or to create a new password pad, and so on...) will always remain the same. Disregarding which pad we're currently looking at. So there's no point in having to redefine and re-create them every time that the `/etherpad` page re-renders.

By moving each action into its own file this stops from happening. It will be defined once. And React does its magic to re-use them whenever possible, and only re-renders them if really needed.

Let's stick to this for newly created code. 🚀

### DIFF
--- a/pages/etherpad/[[...roomId]].js
+++ b/pages/etherpad/[[...roomId]].js
@@ -13,10 +13,13 @@ import { ServiceSubmenu } from '../../components/UI/ServiceSubmenu';
 import BinIcon from '../../assets/icons/bin.svg';
 import ClipboardIcon from '../../assets/icons/clipboard.svg';
 import { ServiceTable } from '../../components/UI/ServiceTable';
-import Form from '../../components/UI/Form';
 import LoadingSpinnerInline from '../../components/UI/LoadingSpinnerInline';
 import LoadingSpinner from '../../components/UI/LoadingSpinner';
 import logger from '../../lib/Logging';
+import CreateAnonymousPad from './actions/CreateAnonymousPad';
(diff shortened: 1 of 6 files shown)


pages/etherpad/[[...roomId]].js
```diff
@@ -172,132 +175,11 @@ export default function Etherpad() {
         return room;
     }, [auth, matrix, matrixClient, etherpad]);
 
-    const ActionNewAnonymousPad = ({ callbackDone }) => {
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:175]: Not actually removing any of this code. These 4 functions were just moved to their own files, each.

marcel-klasse (approved):
Nice!
Only question is whether we should merge this before #11 and adapt `spacedeck`, or merge #11 first, then merge this one, then adapt `spacedeck`.

PR #42, opened 2023-08-23
Invite users to service rooms 

aofn:
Adds a component which opens an "invite user to matrix room" dialogue.

I've used [react-modal](https://github.com/reactjs/react-modal) for now, to open the dialogue. This way the component can be used within any component at any place.  

The styling is very basic and your typical overlayed window style. We could also change this to float right and be more like an insert or cover the width of the `iframeView`. But I think this will do for now.

- [x] add basic modal view
- [x] ability to search matrix user base
- [x] invite selected user to room
- [x] disable submit button until user was selected from search results
- [x] show feedback and close modal on success 
- [x] clean up component and create subcomponents were necessary
- [x] setAppElement for screen readers
- [x] add to etherpad
- [x] add to spacedeck


components/UI/InviteUserToMatrixRoom.js
```diff
@@ -0,0 +1,192 @@
...
+    const debouncedFetchUsersForContributorSearch = useCallback(debounce((val) => fetchUsersForContributorSearch(val), 300), []);
+
+    const fetchUsersForContributorSearch = useCallback(async (a) => {
+        setIsFetchingSearchResults(true);
+        try {
+            const users = await matrixClient.searchUserDirectory({ term: a });
+            // we only update the state if the returned array has entries, to be able to check if users a matrix users or not further down in the code (otherwise the array gets set to [] as soon as you selected an option from the datalist)
+            users.results.length > 0 && setSearchResults(users.results);
+        } catch (err) {
+            console.error('Error whhile trying to fetch users: ' + err);
```
fnwbr, line 94: No `console.` anymore please; use `logger` instead

components/UI/InviteUserToMatrixRoom.js
```diff
@@ -0,0 +1,192 @@
...
+                setUserFeedback(<ErrorMessage>{ err.data?.error }</ErrorMessage>);
+                await new Promise(() => setTimeout(() => {
+                    clearInputs();
+                }, 3000));
+
+                return;
+            });
+
+        // if everything is okay, we let the user know and exit the modal view.
+        setUserFeedback('✓ ' + validUserObject.displayName + t(' was invited and needs to accept your invitation'));
```
fnwbr, line 122: `validUserObject.displayName` is empty when I invite someone by manually putting in their Matrix ID without using the search-autocomplete:
image

components/UI/InviteUserToMatrixRoom.js
```diff
@@ -0,0 +1,192 @@
...
+import CloseIcon from '../../assets/icons/close.svg';
+import ErrorMessage from './ErrorMessage';
+
+if (typeof window !== 'undefined') Modal.setAppElement(document.body);
+
+const Header = styled.header`
+  display: grid;
+  grid-template-columns: 1fr auto;
+  margin-bottom: calc(var(--margin) * 2);
+
```
fnwbr, line 37: Empty line

components/UI/InviteUserToMatrixRoom.js
```diff
@@ -0,0 +1,192 @@
...
+`;
+
+const CloseButton = styled(TextButton)`
+  /* unset globally defined button styles; set height to line-height */
+  width: unset;
+  height: calc(var(--margin) * 1.3);
+  padding: unset;
+  background-color: unset;
+  border: unset;
```
fnwbr, line 46: Don't `TextButton` components have unset the border and backgrounds already?

components/UI/InviteUserToMatrixRoom.js
```diff
@@ -0,0 +1,192 @@
...
+
+    return <>
+        <button title={t('Invite a user to ' + name)} onClick={handleClick}>
+            <UserAddIcon fill="var(--color-foreground)" />
+        </button>
+        { isInviteDialogueOpen && (
+            <Modal
+                isOpen={isInviteDialogueOpen}
+                onRequestClose={() => setIsInviteDialogueOpen(false)}
+                contentLabel="Invite Users"
```
fnwbr, line 158: Please translate

fnwbr (changes requested):
1. Why do we show parenthesis in the dropdown, but not in the input later on? I like it better with parenthesis like in the search results. Can we do the same in the input?
    1. image
    2. \*clicking the last row\*
    3. image
2. After clicking one of the rows, there's an unnecessary search request happening, that's always returning an empty array as the result. Can we stop that from happening? I feel like there's no point in that. Or is there?
    image

marcel-klasse:
>     2. After clicking one of the rows, there's an unnecessary search request happening, that's always returning an empty array as the result. Can we stop that from happening? I feel like there's no point in that. Or is there?

The problem is really datalist not having events... As soon as the user selects an entry from the datalist, the input changes and therfore starts a new request. So we would have to cycle through the last search results to see if the input matches any of the results. 

I went ahead and swapped the datalist for a `ServiceTable` component in 1c935a0813d7890ffc0dd3d83e15fb2e63657c74. Makes everything a little easier to handle on the code side.

fnwbr:
After clicking "Invite x to y" I get this:
image

Console is empty; no network requests happening

fnwbr:
Also, after inviting someone, closing the popup, and then opening the popup again, the button still says "Invite x to y"
image and the button is clickable, when it shouldn't be.

Also... the copy reads weird. Let's slim it down. But I can also do that in my final touches...

fnwbr:
image

Can we turn the line saying "florian" into something clickable? Like an `<a>` or a `<button>`?

Also: this looks bad... can we do something about the comma? Something that looks better? What about `Real Name (@username:matrix.host)` maybe? @andirueckel

components/UI/Datalist.js
```diff
@@ -0,0 +1,114 @@
...
+                ref={inputRef}
+                disabled={isLoading}
+            />
+            { options.length > 0 && isOpen && (
+                <ServiceTable>
+                    { options.map((option, index) => (
+                        <ServiceTable.Row
+                            key={index}
+                            selected={selectedIndex === index}
+                            onClick={() => handleSelect(option)}>
```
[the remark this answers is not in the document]
fnwbr: So... as I mentioned somewhere else previously on this Pull Request: I think most semantic would be a `<button>` inside the `<ServiceTable.Cell`... I'm sure a `<TextButton>` would look just fine.

components/UI/InviteUsersToMatrixRoom/UserListEntry.js
```diff
@@ -0,0 +1,33 @@
...
+    };
+
+    return <ServiceTable.Row>
+        <ServiceTable.Cell>{ user.display_name } ({ user.user_id })</ServiceTable.Cell>
+        <ServiceTable.Cell>
+            <TextButton
+                title={t('invite {{user}} to join {{room}}', { user: user.display_name, room: roomName })}
+                onClick={handleClick}
+                disabled={isInviting}
+            >{ isInviting ? <LoadingSpinnerInline /> || '✓' : <UserAddIcon fill="var(--color-foreground)" /> }
```
fnwbr, line 27: This is a weird inline `if`... when will we get to see the `'✓'` for example?
I think you're just missing some code logic here; or it's copy-🍝 and we don't actually need the `'✓'` here.

components/UI/Datalist.js
```diff
@@ -0,0 +1,114 @@
...
+
+                        </ServiceTable.Row>
+                    )) }
+                </ServiceTable>
+            ) }
+        </div>
+    );
+}
+
+export default Datalist;
```
fnwbr, line 114: When there's not a lot of top-level stuff happening in a file -- like in this one -- you can do this all in one line:

```js
export default function Datalist({...}) {
}
```

This is also what the React documentation recommends: https://react.dev/learn/your-first-component#defining-a-component

Please adjust the other files accordingly. (`DefaultModal`, `UserListEntry`)

Going to see if I can get ESLint to remind us of this in the future.

components/UI/Datalist.js
```diff
@@ -0,0 +1,114 @@
...
+    };
+
+    const handleKeyDown = (e) => {
+        // Handle keyboard navigation
+        if (e.key === 'ArrowDown') {
+            e.preventDefault();
+            if (!isOpen) {
+                setIsOpen(true);
+            } else {
+                const newIndex = Math.min(selectedIndex + 1, options.length - 1);
```
fnwbr, line 43: No need to define a `const` really here, or is there?
If you move this right into the line below, then there's one less variable for our JavaScript memory to keep track of.

Same goes for the lines below.

components/UI/Datalist.js
```diff
@@ -0,0 +1,114 @@
+import React, { useState, useRef } from 'react';
+
+import { ServiceTable } from './ServiceTable';
+
+/**
+ * An input component that functions as a datalist and can be controlled with arrow keys and mouse clicks.
+ *
+ * @component
+ * @param {string[]} options - An array of Objects for the datalist.
```
fnwbr, line 9: If this is an "array of objects", then it should be documented as `@param {Object[]} options`.

~~But instead I'd like to suggest a simplification of this `Datalist` component:
Instead of passing an unnecessary complex array of objects just because this was the use case you had on hand when designing this, let's think of this from the `Datalist` point of view. We, as `Datalist`, don't really know what's in that array of objects; this is the reason why you ended up with this `keysToDisplay` property (and the inferred logic).~~

~~Let's think of it the other way around and look at what `Datalist` actually needs:~~
1. ~~A string to display in the list~~
2. ~~Some sort of identifier, ID, index, key to report back to the consumer component after an entry was selected~~

~~Coming from that, this should actually only need what you said in the very beginning in the documentation: an **array of strings**. Something like:~~
image

~~And once an entry gets selected, you just report back the key. So when `@andi:medienhaus.dev` gets selected, you call `onSelect(1)`. Since the consumer component passed this `options` array, it obviously _knows_ that `1` is `@andi` in this case.~~

~~And -~~

After writing all of the above I realized, that we want to show more than just a string. It's actually a table with different columns... and yes; in that case it makes sense to do what you're doing here.

At the same time I'm unhappy with the user experience. When typing in user names the table "shakes" a lot. Readability is hard when the widths of columns in a table start shifting around with every key you type.

For example in this case, I think the text in the right column could be aligned to the right. To ensure fixed focus points for the eye.

Can you maybe work on polishing this once again with @andirueckel? I'll leave some more UX-related stuff in the final review comment.

public/locales/de/invitationModal.json
```diff
@@ -0,0 +1,6 @@
+{
+  "was invited and needs to accept your invitation": "wurde eingeladen und muss die Einladung annehmen",
+  "Error while trying to fetch users. ": "Error beim Abruf der User.",
+  "Invite users to": "Lade User ein zu",
+  "invite {{user}} to {{room}}": "Lade {{user}} zu {{room}}"
```
fnwbr, line 5: > Lade Florian zu Pad

?

public/locales/de/invitationModal.json
```diff
@@ -0,0 +1,6 @@
+{
+  "was invited and needs to accept your invitation": "wurde eingeladen und muss die Einladung annehmen",
```
fnwbr, line 2: Please put the user / name variable inside the string. In the case of German & English it might work to assume that it always comes at the beginning of the sentence.
But in other languages that might not be the case. So it's always good practice to include everything.

fnwbr (changes requested):
@andirueckel Actually, can you help out polishing the translation? You can make edits right on this branch.

components/UI/Datalist.js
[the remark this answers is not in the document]
marcel-klasse: problem is that neither `<a />` nor `<button />` should be a child of `<table />` 🤔

andirueckel:
@marcel-klasse I tested your latest changes locally, and the quasi _datalist_ interaction seems to work fine—besides the very short color changes from regular to faded back to regular after mouse-click *checking*/*unchecking* a list item which you already mentioned during our previous tests.

Let’s deploy the pull-request branch to our `develop` environment and do some more testing with our respective accounts across the instance and with supported services next week.

fnwbr:
please just re-request a review from me once you're fully happy :)

fnwbr:
Please optimize for mobile viewports.

components/UI/InviteUsersToMatrixRoom/index.js
```diff
@@ -0,0 +1,126 @@
...
+        // if everything is okay, we let the user know and exit the view.
+        successAmount > 0 && setUserFeedback('✓ ' + successAmount + ' ' + t('{{user}} invited and {{needs}} to accept your invitation', { user: successAmount > 1 ? 'users were' : 'user was', needs: successAmount > 1 ? 'needs' : 'need' }));
+        await new Promise(() => setTimeout(() => {
+            clearInputs();
+            if (onSuccess && successAmount === selectedUsers.length) onSuccess();
+        }, 3000));
+    };
+
+    return <ActionWrapper>
+        <h3>{ t('Invite users') }</h3>
```
fnwbr, line 109: Missing translations for this one, and even more. The button also says "Invite" in the German version.

fnwbr (changes requested):
I cleaned up the PR a bit, also added a line to filter out the user who is currently inviting from the search results.

Please double-check the translations. Also the "needs" and "need" needs to be swapped.

components/UI/InviteUsersToMatrixRoom/index.js
marcel-klasse: I've added a <Trans /> component for the more complicated translation now with commit b62f64ab9e8090de341f8b37971e126b1c7137ff. Improves readability and also makes potential future translations easier. Had to add an additiona `en/invitationModal.json` file for this, tho.

components/UI/InviteUsersToMatrixRoom/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+        if (errors.length !== 0) {
+            // if something went wrong we display the errors and clear all inputs
+            setErrorFeedback(errors);
+        }
+
+        const successAmount = selectedUsers.length - errors.length;
+
+        // if everything is okay, we let the user know and exit the view.
+        successAmount > 0 && setUserFeedback(<Trans t={t} i18nKey="invitedUser" count={successAmount}>{ { successAmount } } user was invited and needs to accept your invitation</Trans>);
```
fnwbr, line 104: A cleaner way of combining translations with feedback messages would be to set the state here, and then somewhere down use the `<Trans>` component (or `t()` function respectively). For example, you could rename the state to be called something like `invitedUsersCount` and then at the bottom, where we render, instead of

```jsx
{ userFeedback && _.isEmpty(errorFeedback) ? <div>{ userFeedback }</div> :
```

... you would ...
```jsx
{ invitedUsersCount && <Trans ... count={invitedUsersCount}>bla bla bla</Trans> }
```

This would also allow to switch languages _while_ the message is showing up.
Right now we're writing a translated string into the state, which is not really what the state is made for.

public/locales/de/invitationModal.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Invite users to {{name}}": "Benutzer*innen zu {{name}} einladen",
+  "invitedUser_one": "{{ count }} Benutzer*in wurde eingeladen und muss die Einladung annehmen",
+  "invitedUser_other": "{{ count }} Benutzer*innen wurden eingeladen und müssen die Einladung annehmen",
+  "Error while trying to fetch users. ": "Error beim Abruf der User.",
```
fnwbr, line 5: this is a really _meh_ translation... 😕

fnwbr: Also: Why do we have a period here? But no periods at the end of sentences above?

public/locales/de/invitationModal.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Invite users to {{name}}": "Benutzer*innen zu {{name}} einladen",
```
fnwbr, line 2: How about "Jemanden" instead of "Benutzer*innen"?

fnwbr (approved):
I have some remarks, but I'll go ahead and merge this as is. We can refine on `main`.

Issue #30, opened 2023-07-26
Remove hardcoded Etherpad & MyPads urls


fnwbr:
Since the `/p/` might not actually be present, let's make sure we use config variables instead of doing some search & replace magic.

https://github.com/medienhaus/medienhaus-spaces/blob/5129beb85dd1b7e2717d23827f3fa4b94d3488a4/pages/etherpad/%5B%5B...roomId%5D%5D.js#L235-L237

PR #49, opened 2023-08-29
Display invitations to services in /dashboard

marcel-klasse:
Adds a basic dashboard displaying invitations for each service declared in the config.
Also fetches `dev.medienhaus.meta` for invitations.

- [x] display source (service) of invitation
- [x] display user name who sent invitation

Known Issues:

-  There's a problem with `matrix.invites` showing already handled invitations. To reproduce: Accept invitation for a pad or sketch (click cancel when asked if you want to be redirected) -> reload page.


components/layouts/default.js
```diff
@@ -10,7 +10,6 @@ export const Main = styled.main`
   @media ${breakpoints.tabletAndAbove} {
     grid-row: 1/4;
     grid-column: 2;
-    max-width: 55ch;
```
fnwbr, line 13: image
image

components/layouts/partials/navigation.js
```diff
@@ -26,6 +28,8 @@ const List = styled.ul`
 
 export default function Navigation({ closeNavigation }) {
     const auth = useAuth();
+    const matrix = useMatrix(auth.getAuthenticationProvider('matrix'));
```
fnwbr, line 31: Since https://github.com/medienhaus/medienhaus-spaces/pull/44 was merged it's enough to say `useMatrix()` -- there's no parameter that the function accepts anymore 👍

components/layouts/partials/navigation.js
```diff
@@ -26,6 +28,8 @@ const List = styled.ul`
...
+    const invites = matrix.invites;
```
fnwbr, line 32: Is this variable necessary? Or can we just say `matrix.invites.size` down there?

pages/chat/[[...roomId]].js
```diff
@@ -136,7 +136,7 @@ export default function RoomId() {
         };
     });
 
-    const invites = _.sortBy([...matrix.invites.values()], sortRooms);
+    const invites = _.sortBy([...matrix.invites.values()], sortRooms).filter(invite => !invite.meta);
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:139]: Can you add a comment what this `.filter(invite => !invite.meta);` call does "in layman terms"?
Is it there to show all invites, _except_ for the ones that we assume to be part of one application?

Which makes me wonder: Wouldn't normal "chats" that we start from the /spaces app also have a `dev.medienhaus.meta` event in the future? Is this filter accurate then?

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,70 @@
...
+import { ServiceTable } from '../../components/UI/ServiceTable';
+import TextButton from '../../components/UI/TextButton';
+import AcceptIcon from '../../assets/icons/accept.svg';
+import CloseIcon from '../../assets/icons/close.svg';
+import LoadingSpinnerInline from '../../components/UI/LoadingSpinnerInline';
+
+/**
+ * COMPONENT 'DisplayInvitations'
+ *
+ * @TODO
```
fnwbr, line 14: What's left `@TODO`?

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,70 @@
...
+import { useRouter } from 'next/router';
+
+import { ServiceTable } from '../../components/UI/ServiceTable';
+import TextButton from '../../components/UI/TextButton';
+import AcceptIcon from '../../assets/icons/accept.svg';
+import CloseIcon from '../../assets/icons/close.svg';
+import LoadingSpinnerInline from '../../components/UI/LoadingSpinnerInline';
+
+/**
+ * COMPONENT 'DisplayInvitations'
```
fnwbr, line 12: Might as well remove this comment line... it's not really adding any value, or is it?

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,70 @@
...
+                { name }
+            </ServiceTable.Cell>
+            <ServiceTable.Cell>
+                { invite.name }
+            </ServiceTable.Cell>
+            <ServiceTable.Cell title={invite.inviter?.userId}>
+                { invite.inviter?.displayName }
+            </ServiceTable.Cell>
+            <>
+                <ServiceTable.Cell title={t('accept invitation')} onClick={(e) => { handleAccept(e, invite.roomId); }}>
```
fnwbr, line 54: As always: Please add translations for new strings... please always check for this yourselves before asking for a review. Same goes for `decline invitation` below...

pages/dashboard/ServiceInvitations.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
+import getConfig from 'next/config';
+
+import DisplayInvitations from './DisplayInvitations';
+
+/**
+ * COMPONENT 'ApplicationSection'
```
fnwbr, line 7: Please clean up and add some text for what this component is doing... I think it's worth it here.

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+    // const authProviders = getConfig().publicRuntimeConfig.authProviders; // get the authProvider object from the config
+    // const providersWithTemplates = useMemo(() => {
+    //     return _.pickBy(authProviders, provider => _.has(provider, 'templates')); // only return authProviders which have a templates key
+    // }, [authProviders]);
+
+    // functions which interact with matrix server
+    const declineMatrixInvite = async (e, roomId) => {
+        e.preventDefault();
+        console.log('rejecting ' + roomId);
```
fnwbr, line 46: Please use `logger.debug()` instead...

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+import { useAuth } from '../../lib/Auth';
+import { useMatrix } from '../../lib/Matrix';
+import ServiceInvitations from './ServiceInvitations';
+import { ServiceTable } from '../../components/UI/ServiceTable';
+import DisplayInvitations from './DisplayInvitations';
+
+const sortRooms = function(room) {
+    return [
+        room.notificationCount === 0,
```
fnwbr, line 15: Doesn't really make sense to sort them by the notification count --- pending invitations always have a `notificationCount` of `1`, I guess. At least it's always the same.

I think it would make more sense to replace the `_.sortBy([...matrix.invites.values()], sortRooms)` line with something like: `_.sortBy([...matrix.invites.values()], 'name')` instead.

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+`;
+export default function Dashboard() {
+    const auth = useAuth();
+    const matrix = useMatrix(auth.getAuthenticationProvider('matrix'));
+    const MatrixAuthProvider = auth.getAuthenticationProvider('matrix');
+    const { t } = useTranslation('dashboard');
+
+    const invites = matrix.invites;
+    const matrixClient = auth.getAuthenticationProvider('matrix').getMatrixClient();
+    const serviceSpaces = matrix.serviceSpaces;
```
fnwbr, line 35: Can't see this variable being used...

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+`;
+export default function Dashboard() {
+    const auth = useAuth();
+    const matrix = useMatrix(auth.getAuthenticationProvider('matrix'));
+    const MatrixAuthProvider = auth.getAuthenticationProvider('matrix');
+    const { t } = useTranslation('dashboard');
+
+    const invites = matrix.invites;
+    const matrixClient = auth.getAuthenticationProvider('matrix').getMatrixClient();
```
fnwbr, line 34: I think this is getting a little confusing overall... we have too many layers:

1. `useMatrix()` -- which is referring to things & functions inside `Matrix.js`
2. `auth.getAuthenticationProvider('matrix')` -- referring to functions inside `MatrixAuthProvider.js`
3. `auth.getAuthenticationProvider('matrix').getMatrixClient()` -- referring to the actual "raw" `matrix-js-sdk` `MatrixClient`

Do you have an idea how to simplify this?
One thing I can think of would be to never call (2) from anywhere, except for from `Matrix.js`. Which would mean that all functions we've added in `MatrixAuthProvider.js` (like `addSpaceChild`) would live in `Matrix.js` instead.

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+        console.log('rejecting ' + roomId);
+        await matrix.leaveRoom(roomId);
+    };
+
+    const acceptMatrixInvite = async (e, roomId, service, name) => {
+        e.preventDefault();
+        await matrixClient.joinRoom(roomId).catch(() => {
+            return;
+        });
+        await MatrixAuthProvider.addSpaceChild(serviceSpaces[service], roomId);
```
fnwbr, line 55: Sooo... when we accept an invite for e.g. a pad, we add this pad to our "private" Etherpad applications folder? Is that correct? Is this line also being executed for "normal" chats? What does this line do in that case?

pages/dashboard/index.js
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    return (
+        <>
+            <h2>/dashboard</h2>
+
+            { /* { invites.size > 0 && <ApplicationSegment>
```
fnwbr, line 62: Remove?

pages/dashboard/ServiceInvitations.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @TODO
+ *
+ *
+ * @param {String} name — name of the Application
+ * @param {String} id — id of the application
```
fnwbr, line 13: This is not accurate... I can see many more parameters down there.
Also, what is the "id of the application"?

pages/dashboard/ServiceInvitations.js
```diff
@@ -0,0 +1,38 @@
...
+
+const ServiceInvitations = ({ service, id, invitations, acceptMatrixInvite, declineMatrixInvite }) => {
+    const name = getConfig().publicRuntimeConfig.authProviders[service].path || service;
+    const serviceTemplates = getConfig().publicRuntimeConfig.authProviders[service].templates;
+    const serviceInvitations = [...invitations.values()].filter(invite => serviceTemplates.includes(invite.meta?.template)); // filter invitations for the current service
+
+    if (_.isEmpty(serviceInvitations)) return null;
+
+    return (
+
```
fnwbr, line 24: Empty line? Also, no need for the `(` in the line above, or is there?

components/UI/ServiceTable.js
```diff
@@ -42,13 +89,17 @@ const TableCell = styled.td`
 
 `;
 
-export function ServiceTable({ children }) {
+export function ServiceTable({ children, selected, align }) {
```
fnwbr, line 92: When are these being used? Can't find any consumers. Shouldn't `ServiceTable` be just this?
```js
const ServiceTable = Table;
```

lib/Matrix.js
```diff
@@ -138,7 +142,20 @@ function useMatrixProvider(auth) {
...
+            const metaEvent = await hydrateMetaEvent(room.roomId);
+            // we check to see who invited the current user
+            const inviterName = room.getMember(matrixClient.getUserId())?.events?.member?.getSender?.();
+            const inviter = matrixClient.getUser(inviterName);
+            setInvites(setRoom(room.roomId, metaEvent, inviter));
+        }
+
+        if (prevMembership === 'invite') {
+            // in this case we know the invitation was handled by the user and we can delete it from our invitations
+            setInvites(deleteElement(room.roomId));
```
fnwbr, line 158: Isn't line `137` taking care of this already? If not, why not and in which case?

lib/Matrix.js
```diff
@@ -138,7 +142,20 @@ function useMatrixProvider(auth) {
             }
         }
         if ((membership === 'invite')) {
-            setInvites(setRoom(room.roomId));
+            // we need to update the matrixClient room sync so getRooms(roomId) in buildRoomObject doesn't return null
+            await matrixClient.roomInitialSync(room.roomId)
```
fnwbr, line 146: Find it hard to explain this quickly, but as a rule of thumb: Never do any `async` actions inside the event handler functions.

[These functions](https://github.com/medienhaus/medienhaus-spaces/pull/49/files#diff-2f8d4752f5ecbf336e6d4aeb24e15c2483c854d3d0c24b137625b5ad3376d301L186-L188) get executed for each and every single event of those types that can be found in the Matrix. This includes **all events** during the initial sync at the beginning, and this includes every event of those types that happens afterwards when the user has the page open.

To get around this let's think about another way where we can do this: If I understood correctly you need this in order to display detailed information per invitation on the dashboard.
Do we need that detailed information at all times? Or do we only need it in the dashboard?

If it's the latter, could you think of a way of fetching that `inviter` information only when looking at the dashboard?

If this gets too complicated I'm happy to make these changes myself, after you've addressed the other review comments.

next.config.example.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
                     api: 'http://etherpad.localhost/mypads/api',
                     spacesGroupName: '/spaces', // optional, defaults to publicRuntimeConfig.name
                 },
+                templates: ['etherpad-link'],
```
fnwbr, line 18: Is this the correct default? When looking at my rooms from my test instance they all have their template set to `write-link`... I think this is another thing that needs to be addressed with https://github.com/medienhaus/medienhaus-spaces/issues/30 for more consistency.

IMHO this should be just `etherpad`. And udk/spaces knows how to interpret `etherpad`. No matter how the sysadmin/developer decided to call that "app". And `etherpad` will also always be the known by the codebase, as things are called `EtherpadAuthProvider`, `/pages/etherpad/<...>`, and so on here.
This would work **without** defining a template. The template would just always be the exact same as the technical name of the "service".

fnwbr (changes requested):

components/layouts/default.js
marcel-klasse: i will just @andirueckel here 🙃

pages/chat/[[...roomId]].js
Comment (marcel-klasse) [reply pages/chat/[[...roomId]].js:139]: I've added a comment.

> Which makes me wonder: Wouldn't normal "chats" that we start from the /spaces app also have a dev.medienhaus.meta event in the future? Is this filter accurate then?

Yeah this is sort of an ongoing discussion, whether we need a `dev.medienhaus.meta` event for chat rooms or not. I personally don't see the advantage of "spamming" chat rooms with unnecessary events, or to be able to verify the origin? (as in coming from the /spaces app) but I don't really mind either way.

components/layouts/default.js
andirueckel: The highlighted `max-width`  was removed from the “default layout” on purpose.

We should not, by default, limit the width of our `main` element; this *was* done in another prototype with very different layout and structure. If we want to limit the widths of the sections shown in the above screenshots, then we should do so in their respective components. I’d argue, though, that we might want to adapt the layout logic/structure per se as we made many changes to [medienhaus-spaces](https://github.com/medienhaus/medienhaus-spaces).

pages/chat/[[...roomId]].js
Comment (andirueckel) [reply pages/chat/[[...roomId]].js:139]: > Which makes me wonder: Wouldn't normal "chats" that we start from the /spaces app also have a dev.medienhaus.meta event in the future? Is this filter accurate then?

No, we should not add any `dev.medienhaus.meta` events to **chat** rooms.

Chat rooms are considerably well defined, and _none of our business_; we _use_ chat rooms, but we don’t want to re-invent chat rooms or break any compatibility with, well, *any* matrix chat room from any matrix chat client, like for example [element-web](https://github.com/vector-im/element-web).

pages/dashboard/DisplayInvitations.js
marcel-klasse: updating the JSDoc 😁

lib/Matrix.js
marcel-klasse: ah yeah, I think this was left over from me trying to fix the weird issue with invitations reappearing.

lib/Matrix.js
marcel-klasse: changed it to fetch the information within dashboard.

If we want to display invitations for individual services on the actual service page, as well as on the dashboard it will probably make sense to fetch this inforamtion directly when we receive an invitation. But then let's cross that bridge when we get to it 🙂

components/layouts/default.js
fnwbr: I'm cool with either way. I'm just not cool with how the `/login` page is looking now. Feel free to implement whatever you prefer, or bring back the `max-width` for now and propose a more general layout change in a separate PR.

fnwbr (changes requested):

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,80 @@
...
+ * @param e - event triggered by the button
+ * @param roomId - matrix roomId
+ *
+ * @returns {React.ReactComponent}
+*/
+
+const DisplayInvitations = ({ invite, name, service, acceptMatrixInvite, declineMatrixInvite }) => {
+    const { t } = useTranslation('dashboard');
+    const [isAcceptingInvite, setIsAcceptingInvite] = useState(false);
+    const [isDecliningInvite, setisDecliningInvite] = useState(false);
```
fnwbr, line 34: capital `I` in `setisDecliningInvite` please

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,80 @@
...
+        setisDecliningInvite(true);
+        await declineMatrixInvite(e, roomId);
+        setisDecliningInvite(false);
+    };
+
+    const handleAccept = async (e, roomId) => {
+        setIsAcceptingInvite(true);
+        await acceptMatrixInvite(e, roomId, service);
+        setIsAcceptingInvite(false);
+        if (confirm('You’ve successfully accepted the invitation!\n\nWould you like to be redirected to the newly accepted ' + name + ' item?')) {
```
fnwbr, line 47: Missing translations...

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,80 @@
...
+
+const DisplayInvitations = ({ invite, name, service, acceptMatrixInvite, declineMatrixInvite }) => {
+    const { t } = useTranslation('dashboard');
+    const [isAcceptingInvite, setIsAcceptingInvite] = useState(false);
+    const [isDecliningInvite, setisDecliningInvite] = useState(false);
+    const router = useRouter();
+
+    const handleDecline = async (e, roomId) => {
+        setisDecliningInvite(true);
+        await declineMatrixInvite(e, roomId);
```
fnwbr, line 39: Instead of forwarding the `e` only to prevent the default click handler logic, please do that here.
As a rule of thumb: Catch those default behaviors always as close as possible to the original sending element. In this case it's the table cell. And not something that's defined in the file where you're currently calling `e.preventDefault()`. This will also allow you to not having to pass the `e` down the call stack all the way.

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,80 @@
...
+        setisDecliningInvite(false);
+    };
+
+    const handleAccept = async (e, roomId) => {
+        setIsAcceptingInvite(true);
+        await acceptMatrixInvite(e, roomId, service);
+        setIsAcceptingInvite(false);
+        if (confirm('You’ve successfully accepted the invitation!\n\nWould you like to be redirected to the newly accepted ' + name + ' item?')) {
+            router.push(`${name}/${roomId}`);
```
fnwbr, line 48: Instead of _assuming_ that this path pattern always works, can we make the corresponding consumer define it? Something I can think of, for example, would be to return it in the `acceptMatrixInvite` function like this:
```js
const forwardingUrl = await acceptMatrixInvite(e, roomId, service);

...

if (...) { router.push(forwardingUrl); }
```

pages/dashboard/DisplayInvitations.js
fnwbr: (Same goes for `acceptMatrixInvite` below)

fnwbr (changes requested):

pages/chat/[[...roomId]].js
Comment (fnwbr) [reply pages/chat/[[...roomId]].js:147]: Cool, agreed then. 👍  Let's just add a short comment on what the filtering is supposed to do. Feel free to leave a link to _this_ very discussion here on GitHub inside the code.

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,80 @@
...
+            <ServiceTable.Cell title={invite.inviter?.userId}>
+                { invite.inviter?.displayName }
+            </ServiceTable.Cell>
+            <>
+                <ServiceTable.Cell title={t('accept invitation')} onClick={(e) => { handleAccept(e, invite.roomId); }}>
+                    <TextButton disabled={isDecliningInvite || isAcceptingInvite}>
+                        { isAcceptingInvite ? <LoadingSpinnerInline /> : <AcceptIcon /> }
+                    </TextButton>
+                </ServiceTable.Cell>
+                <ServiceTable.Cell title={t('decline invitation')} onClick={(e) => {handleDecline(e, invite.roomId);}}>
```
fnwbr, line 69: A quick detour into accessibility land: As a rule of thumb you don't want to add `onClick` handlers to anything but a `<button>` (or `<a>` elements, when changing the URL)...
The short solution would be to add this to the `<TextButton>` element below instead.

If you want to read up on this:
- https://github.com/reactjs/react-a11y/blob/master/docs/rules/onclick-uses-role.md (and the W3 article linked at the bottom)
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#accessibility_concerns

fnwbr (changes requested):

pages/dashboard/index.js
marcel-klasse: maybe i'll change this in a separate PR? the same would go for `removeSpaceChild`

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,84 @@
...
+        setIsDecliningInvite(false);
+    };
+
+    const handleAccept = async (e, roomId) => {
+        e.preventDefault();
+        setIsAcceptingInvite(true);
+        const forwardingUrl = await acceptMatrixInvite(roomId, service);
+        setIsAcceptingInvite(false);
+        if (forwardingUrl) {
+            if (confirm('You’ve successfully accepted the invitation!\n\nWould you like to be redirected to the newly accepted {{name}} item?', { name: path })) {
```
fnwbr, line 48: I'm not sure if I like this... let's assume I have 10 invitations. So instead of being able to quickly accept/reject them, I have to confirm or deny this after every step.

It's more like this: either I want to be forwarded to the item I just accepted the invitation for, or I don't and I want to stay on my dashboard. But if I choose the latter, why would I decide otherwise after accepting/rejecting the next invitation?

I'm not sure what to suggest but removing this `confirm()` altogether.
Or somehow memorizing if the user said "no, don't forward me" and then not doing so for all subsequent accepts.

What do you think from an UX POV @andirueckel?

fnwbr: Also, something's not working:
image

fnwbr (changes requested):
1. After accepting a `/chat` invitation I got redirected to this:
  image
2. Why does this say `write`? Shouldn't there be a `/` in front? I'm using the same config as `next.config.example.js`.
  image

pages/dashboard/DisplayInvitations.js
andirueckel: > It's more like this: either I want to be forwarded to the item I just accepted the invitation for, or I don't and I want to stay on my dashboard. But if I choose the latter, why would I decide otherwise after accepting/rejecting the next invitation?

@fnwbr @marcel-klasse Let’s replace the `confirm()` with `alert()` and remove the ”Would you like to be redirected …” paragraph. After having accepted the invitation for/to an item, let’s disable the respective accept/decline buttons and make the item’s display name a proper hyperlink.

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,78 @@
...
+            <ServiceTable.Cell>
+                { path }
+            </ServiceTable.Cell>
+            <ServiceTable.Cell>
+                { invite.name }
+            </ServiceTable.Cell>
+            <ServiceTable.Cell title={invite.inviter?.userId}>
+                { invite.inviter?.displayName }
+            </ServiceTable.Cell>
+            <>
```
fnwbr, line 61: Why is this empty element here?

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,78 @@
...
+                </ServiceTable.Cell>
+                <ServiceTable.Cell title={t('decline invitation')}>
+                    <TextButton onClick={(e) => {handleDecline(e, invite.roomId);}} disabled={isDecliningInvite || isAcceptingInvite}>
+                        { isDecliningInvite ? <LoadingSpinnerInline /> : <CloseIcon /> }
+                    </TextButton>
+                </ServiceTable.Cell>
+            </>
+
+        </ServiceTable.Row>
+
```
fnwbr, line 75: Empty line

pages/dashboard/DisplayInvitations.js
```diff
@@ -0,0 +1,78 @@
...
+                        { isAcceptingInvite ? <LoadingSpinnerInline /> : <AcceptIcon /> }
+                    </TextButton>
+                </ServiceTable.Cell>
+                <ServiceTable.Cell title={t('decline invitation')}>
+                    <TextButton onClick={(e) => {handleDecline(e, invite.roomId);}} disabled={isDecliningInvite || isAcceptingInvite}>
+                        { isDecliningInvite ? <LoadingSpinnerInline /> : <CloseIcon /> }
+                    </TextButton>
+                </ServiceTable.Cell>
+            </>
+
```
fnwbr, line 73: Empty line

pages/dashboard/ServiceInvitations.js
```diff
@@ -0,0 +1,38 @@
...
+    return _.map(serviceInvitations, (invite) => {
+        return <DisplayInvitations
+            key={invite.roomId}
+            service={service}
+            path={path}
+            invite={invite}
+            declineMatrixInvite={declineMatrixInvite}
+            acceptMatrixInvite={acceptMatrixInvite} />;
+    })
+
```
fnwbr, line 34: Empty line... and the line below can go to line 33

fnwbr (changes requested):

public/locales/de/_defaults.json
```diff
@@ -8,5 +8,9 @@
   "Save changes": "Änderungen speichern",
   "Select action": "Aktion wählen",
   "What would you like to do?": "Was möchtest Du tun?",
-  "{{error}}, please wait {{time}} seconds.": "{{error}}, bitte warte {{time}} Sekunden."
-}
\ No newline at end of file
+  "{{error}}, please wait {{time}} seconds.": "{{error}}, bitte warte {{time}} Sekunden.",
+  "Invitations": "Einladungen",
+  "From": "Von",
+  "Accept": "Annehemen",
```
fnwbr, line 14: image

fnwbr (changes requested):

fnwbr (approved):

Issue #39, opened 2023-08-17
PR#23 followup: Rename `EtherpadsAuthProvider` to `MyPadsAuthProvider`


fnwbr:
As discussed per https://github.com/medienhaus/medienhaus-spaces/pull/23#issuecomment-1644568734 this issue is merely a reminder to not forget about what we decided.

---

- [x] Rename `EtherpadsAuthProvider` to `MyPadsAuthProvider`
- [x] Test `router.push()` forwardings are still working
  - [x] Fix these warnings:
    <img width="1407" alt="image" src="https://github.com/medienhaus/medienhaus-spaces/assets/706419/7693c565-b163-4561-b272-9ffc498757b7">
  - [x] Replace all `router.push('/write...')` mentionings with `router.push('/etherpad...')` (or something else that makes sense; but with what we agreed upon above `/write` just does not make sense anymore)

PR #51, opened 2023-08-31
Cleanup `etherpad`

marcel-klasse:
Fixes issues from  #39


fnwbr (approved):
While reviewing this I had some more ideas, but this is an improvement as is 👍

PR #55, opened 2023-09-06
/explore 

marcel-klasse:


fnwbr (changes requested):
1. Create a new pad
2. Open the pad
3. Where is the "Invite users" icon?

Screenshot 2024-01-31 at 12 13 45

PR #60, opened 2023-09-12
Introduce new medienhaus-spaces specific `config.js`

fnwbr:
This moves everything that is **not** related to our own project to other files, so that users who want to use `medienhaus-spaces` only look at what will actually matter to them.

One single `config.js` file (or `config.example.js`), which only holds information about configuring this app.

Everything that's specific to how we use Next.js or Webpack happens in `next.config.js`, but this will most likely remain untouched from people who deploy the `medienhaus-spaces` application.


config.example.js
```diff
@@ -0,0 +1,24 @@
...
+        },
+    },
+    contextRootSpaceRoomId: '!gB.....Ewlvdq:matrix.org',
+    account: {
+        allowAddingNewEmails: true,
+    },
+    chat: {
+        pathToElement: '//localhost/element',
+    },
+};
```
fnwbr, line 24: So basically -this- file is going to be the new **medienhaus-spaces** config file that we ask developers and administrators to look at when wanting to configure our app.

next.config.js
```diff
@@ -0,0 +1,29 @@
...
+        if (!dev || isServer) return config;
+
+        config.plugins.push(new StylelintPlugin({
+            files: ['assets/*.css', 'components/**/*.js', 'pages/**/*.js'],
+            failOnError: false,
+        }));
+
+        return config;
+    },
+};
```
fnwbr, line 29: ... and this file is where we can put all of our NextJS-specific workarounds and whatabouts like URL rewrites, webpack adjustments and so on.

webpack.config.js
```diff
@@ -1,12 +0,0 @@
...
-module.exports = (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
-    if (!dev || isServer) return config;
-
-    config.plugins.push(new StylelintPlugin({
-        files: ['assets/*.css', 'components/**/*.js', 'pages/**/*.js'],
-        failOnError: false,
-    }));
-
-    return config;
-};
```
fnwbr, line 12: This file is not needed anymore as the content lives inside of `next.config.js` now.

andirueckel (approved):
Love it. 👏

PR #66, opened 2023-09-20
Kick users from room

marcel-klasse:
Adds functionality to kick users from a room.
The user needs a higher power level than the user they want to kick.


marcel-klasse:
basically duplicate of #56

PR #68, opened 2023-09-20
storing accesstokens in matrix user accountdata instead of localstorage

robertschnuell:
this PR creates an workaround for some multi device problems. As `mypads` is not capable to create multiple session specific accesstokens this workaround is needed. Instead of that each session creates an own login process for each authprovider, this accesstokens get stored instead in the logged-in [matrix] user accountdata -> `dev.medienhaus.spaces.accesstokens` accountdata stateevent. If a valid token for a specific authprovider is stored in this statevent the client tries to use this instead of creating a new one. If it is not working, the client updates the `dev.medienhaus.spaces.accesstokens` accountdata stateevent correspondingly. 
Additionally to the `signin` and `signout` functions each authprovider needs from now on as well as `addToken` and `validateToken` functions. This is already implemented as an example in the etherpad authprovider.

### DIFF
--- a/lib/Auth.js
+++ b/lib/Auth.js
@@ -66,21 +66,51 @@ function useAuthProvider() {
 
         await fetchUserProfileAndStartClient();
 
-        // Loop through all existing authentication providers and sign in
-        Object.entries(activeAuthentications).forEach(([type, authentication]) => {
-            // Skip the default Matrix auth provider
-            if (type === 'matrix') return;
-            authentication.signin(username, password).catch(() => { });
-        });
+        await getAccessTokensFromAccountData(username, password);
 
         return matrixAuthenticationResponse;
+    };
+
+    const getAccessTokensFromAccountData = async (username, password) => {
+        // we will use `getAccountDataFromServer` instead of `getAccountData` as we can nnot tell for sure if the sync is completly done and dont want to wait for that. So potentially this call is avoidable but it is more managable to handle it like this.
+        const accessTokens = await getAuthenticationProvider('matrix').getMatrixClient().getAccountDataFromServer('dev.medienhaus.spaces.accesstokens');
+        let dataChanged = false;
+
+        if (!accessTokens) {
+            // fetch did not worked or this accountData state event is not set yet. In either way we want to create it.
+            // First we need to login all of the Services
+
+            dataChanged = true;
+        } else {
+            for await (const [type, authentication] of Object.entries(activeAuthentications)) {
+                // Skip the default Matrix auth provider
+                if (type === 'matrix') continue;
+
+                if (accessTokens[type]) {
+                    authentication.addToken(accessTokens[type]);
+                    if (await authentication.validateToken()) {
+                        // everything seems to work out as expected we will cut here
+                        continue;
+                    }
+                }
+                // if the stored token does not work anymore then we will execute a login procedure and store the new generated accesstoken
+                if (! await authentication.validateToken().catch(() => { })) {
+                    dataChanged = true;
+                    accessTokens[type] = await authentication.signin(username, password).catch(() => { });
+                }
+            }
+        }
+
+        if (dataChanged) {
+            // write accessTokensToAccountData
+            await getAuthenticationProvider('matrix').getMatrixClient().setAccountData('dev.medienhaus.spaces.accesstokens', accessTokens);
+        }
     };
 
     const signout = useCallback(async () => {
         window.localStorage.removeItem('medienhaus_access_token');
         window.localStorage.removeItem('medienhaus_user_id');
         window.localStorage.removeItem('medienhaus_hs_url');
-        window.localStorage.removeItem('write_access_token');
 
         // remove localStorage items we've previously set
         window.localStorage.removeItem('mx_access_token');
@@ -106,6 +136,7 @@ function useAuthProvider() {
 
     const fetchUserProfileAndStartClient = useCallback(async () => {
         const profileInfoResponse = await getAuthenticationProvider('matrix').getMatrixClient().getProfileInfo(getAuthenticationProvider('matrix').getMatrixClient().getUserId());
+
         if (!profileInfoResponse) {
             return signout();
         }
--- a/lib/auth/MyPadsAuthProvider.js
+++ b/lib/auth/MyPadsAuthProvider.js
@@ -4,14 +4,33 @@ import getConfig from 'next/config';
 // @TODO: If a user logs in with a different device, the saved auth token becomes invalid and they need to login again
 
 class MyPadsAuthProvider {
-    constructor(configuration) {
+    constructor(configuration, token) {
         // serverPads will be populated with pads from the mypads server
         this.serverPads = {};
         this.url = configuration.myPads?.api;
         // name of the folder the application stores pads in. defaults to application name
         this.spacesGroupName = configuration.myPads?.spacesGroupName || getConfig().publicRuntimeConfig.name;
-        if (typeof window !== 'undefined' && window.localStorage.getItem('write_access_token')) {
-            this.token = window.localStorage.getItem('write_access_token');
+
+        if (token) this.token = token;
+    }
+
+    addToken(token) {
+        this.token = token;
+    }
+
+    async validateToken() {
+        if (!this.token) return false;
+        // we will execute a authprovider specific dummy call with requires auth to check if the token which we got stored is still valid. This methode needs to be implemented in each individual auth provider same as `signin` and `signout`
+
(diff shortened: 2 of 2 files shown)


marcel-klasse (approved):

lib/auth/MyPadsAuthProvider.js
```diff
@@ -4,14 +4,33 @@ import getConfig from 'next/config';
...
+
+    addToken(token) {
+        this.token = token;
+    }
+
+    async validateToken() {
+        if (!this.token) return false;
+        // we will execute a authprovider specific dummy call with requires auth to check if the token which we got stored is still valid. This methode needs to be implemented in each individual auth provider same as `signin` and `signout`
+
+        const tokenValidationCall = await await fetch(this.url + '/userlist' + '?auth_token=' + this.token, {
```
fnwbr, line 25: `await await` ;)

lib/auth/MyPadsAuthProvider.js
```diff
@@ -4,14 +4,33 @@ import getConfig from 'next/config';
 // @TODO: If a user logs in with a different device, the saved auth token becomes invalid and they need to login again
 
 class MyPadsAuthProvider {
-    constructor(configuration) {
+    constructor(configuration, token) {
```
fnwbr, line 7: Is this new parameter ever being used? If not yet, can we remove it for now?

lib/auth/MyPadsAuthProvider.js
```diff
@@ -23,19 +42,15 @@ class MyPadsAuthProvider {
         this.token = login.token;
         this.userId = login.user._id;
         if (!this.token) return;
-        window.localStorage.setItem('write_access_token', this.token);
         await this.syncAllPads(this.token);
+        this.validateToken();
```
fnwbr, line 46: Do you think this call is really necessary?
After all we -just- logged in in the lines above; I think it's safe to assume that the token is valid at this point and we can just return it.

fnwbr (changes requested):
Love it! 3 small requests

lib/Auth.js
```diff
@@ -66,21 +66,45 @@ function useAuthProvider() {
...
+                }
+            } else {
+                // if the stored token does not work anymore then we will execute a login procedure and store the new generated accesstoken
+                dataChanged = true;
+                accessTokens[type] = await authentication.signin(username, password).catch(() => { });
+            }
+        }
+
+        if (dataChanged) {
+            // write accessTokensToAccountData. this function is needed to prevent spamming with eventId's, even if the accesstokens are all still valid and doesnt need to be updated.
```
fnwbr, line 99: To be clear: that last part of the comment

> this function is needed to prevent spamming with eventId's, even if the accesstokens are all still valid and doesnt need to be updated

refers to the `if (dataChanged)` condition, right? As in, you're trying to say that we only write new access tokens to the Matrix account data *if* and *only if* at least one of the access tokens has changed, right?

If I understood that right, can you clarify that in the documentation? So "function" should probably be "condition", and event IDs do not really play a role here, or do they? Is account data based on events?

fnwbr:
Can you also `git merge main` into this branch and update the `SpacedeckAuthProvider` as well? Should be a quick one, now that you've gotten approvals on your way of doing it.

Otherwise we'll have code in `main` that tries to call `signin()` on the `SpacedeckAuthProvider` without receiving an auth token back from it...

lib/Auth.js
robertschnuell: changed

lib/auth/MyPadsAuthProvider.js
robertschnuell: ok, deleted it.

lib/auth/MyPadsAuthProvider.js
robertschnuell: ok, is deleted.

fnwbr (changes requested):
I'm seeing this when I do the following:
image

1. Open private / incognito window
1. Sign in: everything works 👍 everything loads perfectly fine
1. Close private / incognito window (this will reset cookies, local storage & co.)
1. Open a new window and sign into the same account again
1. Go to `/sketch`

But when I hard-refresh

fnwbr:
I guess it has to do with this:
image

fnwbr (changes requested):
Something that also still doesn't work is: renewing the token when it has expired.

To reproduce that try to modify the token for MyPads in your account data via Element (`/devtools`) and then reload medienhaus-spaces.

Right now when doing this for Etherpad there's gonna be that validation call and Firefox says it's erroring because of missing CORS. That's misleading, ignore that CORS part. It's just that the token is invalid and because of that the MyPads API doesn't seem to set all of the needed CORS headers...

robertschnuell:
> Something that also still doesn't work is: renewing the token when it has expired.
> 
> To reproduce that try to modify the token for MyPads in your account data via Element (`/devtools`) and then reload medienhaus-spaces.
> 
> Right now when doing this for Etherpad there's gonna be that validation call and Firefox says it's erroring because of missing CORS. That's misleading, ignore that CORS part. It's just that the token is invalid and because of that the MyPads API doesn't seem to set all of the needed CORS headers...

If we want to do this the clean way, we have to catch errors within each individual authprovider within each individual call which interacts with the given api, as an access token can get invalid at any given time. So in the case that while a session a token gets invalid, an error occurs, which got caught and a new loginprocess for this specific authprovider will be executed. After a new valid authtoken is provided and stores in accountdata of the given user, the authprovider should try to execute the previously failed interaction with the given api once more and proceed where it stopped before, without showing the user any error. So the user will have a seemingly interface interaction and get the expected result. 
But this 'not so edgecase scenario' (as the lifespan of an token within mypads isnt that long by default)  we are not handling at the moment at all. As we are aware of this problem, I would suggest implementing this solution in the near'ish future. The question now remains: should this been done for now within this PR or should this additional check been an own PR? 

In addition should be there a standardized implementation to handling this potential failed api calls, so that it could be implemented in the same way for all auth provides to come? Or is this something we let everyone who is writing an authprovider for a specific application having the choice to handle this problem or ignoring it? (would prefer the first option tho…)

fnwbr:
>  and a new loginprocess for this specific authprovider will be executed

What I'm saying is: no, this is not what your code is doing. Something doesn't work...

lib/Auth.js
```diff
@@ -68,21 +68,45 @@ function useAuthProvider() {
...
+
+        for await (const [type, authentication] of Object.entries(activeAuthentications)) {
+            // Skip the default Matrix auth provider
+            if (type === 'matrix') continue;
+
+            if (accessTokens[type]) {
+                authentication.addToken(accessTokens[type]);
+                const tokenValidation = await authentication.validateToken();
+                if (tokenValidation) {
+                    // everything seems to work out as expected we will cut here
```
fnwbr, line 90: Continuing the discussion from the main thread:

Something is wrong here: `if (tokenValidation) {` ... so _if_ the token is valid, then yes, everything seems to work and we can continue.

But there's no `} else {` in line 92. This code is not doing anything _if_ the token from the Matrix account data is expired or wrong.

fnwbr:
If you try out what I suggested above you'll see what I mean:

> To reproduce that try to modify the token for MyPads in your account data via Element (`/devtools`) and then reload medienhaus-spaces.

fnwbr:
And yes, this is a regression compared to the current version of `main`:

Previously you'd fetch a fresh token from each auth provider every time you sign in.

Now you fetch a fresh token **once** in the very beginning, when there's no token for the given auth provider in your Matrix account data yet. And then you're stuck with that one **forever**.

So as soon as the token expires, you're stuck. Even signing out and signing in will not resolve the issue.

robertschnuell:
I am aware of this and understood was you were wrting. but still my question remains if this should be fixed with the knowleadge that there is a bigger problem within this potential fix and the status quo or to also to fix this one within this PR.

fnwbr:
Nice, the latest change works as intended. For example with Spacedeck:

image

After logging in and finding tokens in `dev.medienhaus.spaces.access` the app tried to ...
1. `401`: ... validate that token by calling `/current` with the old one, only to realize that the token is invalid; so that it ...
2. `201`: ... asked for a new auth token by calling `/session` (with our username and password) and got a fresh one ...
3. `200`: ... to validate the token once again by calling `/current` ⁉️ -- so that ...
4. `200`: ... all actual API requests (e.g. to `/spaces`) will work 👍 

So what I'm unsure about is if the call mentioned in step (3) is actually necessary. I think we could save up on that as we just got a fresh token in the step before. I think we have talked about it during the peer review as well...

PR #69, opened 2023-09-26
Explore: actions within iframe layout

marcel-klasse:
Displays explore actions within the iframe layout


andirueckel (approved):
Needs some layout/spacing refinements, but let’s merge this now and take care of the details later. ✨

PR #71, opened 2023-09-27
Add retry handler for matrix rate limits

aofn:
Uses `retry_after_ms` from the error response to await ratelimiting and retry the failed request.

- [x] fix bug where creating rooms loops and doesn't proceed with adding the room to its parent
- [x] display user feedback while syncing
- [x] add translations
- [x] use retry handler when creating application folder
- [x] use retry handler when creating service folders
- [x] use retry handler for /etherpad

**Testing:**

- try to encounter rate limits. for example create at least 8 sketches directly in spacedeck (not from within spaces! the more the better, since this should trigger ratelimiting by [matrix])
- open /spaces and go to /spacedeck
- /spacedeck will now try to sync all sketches and, if it runs into ratelimits, should automatically wait until it tries again

### DIFF
--- a/lib/auth/MatrixAuthProvider.js
+++ b/lib/auth/MatrixAuthProvider.js
@@ -132,16 +132,6 @@ class MatrixAuthProvider {
                 type: 'm.room.join_rules',
                 content: { join_rule: joinRule }, // can be set to either public, invite or knock
             },
-            {
-                // Spec: https://spec.matrix.org/v1.9/client-server-api/#mspaceparent
-                // Element-Web: https://github.com/matrix-org/matrix-react-sdk/blob/fb1a97be3228ce1341c54a379f8374710b65a0e8/src/utils/space.tsx#L51-L58
-                type: 'm.space.parent',
-                state_key: parentId,
-                content: {
-                    canonical: true,
-                    via: [this.matrixClient.getDomain()],
-                },
-            },
             ],
             power_level_content_override: {
                 // we only want users with moderation rights to be able to do any actions, people joining the room will have a default level of 0.
@@ -161,6 +151,19 @@ class MatrixAuthProvider {
                 users_default: 0,
             },
         };
+
+        if (parentId) {
+            opts.initial_state.push({
+                // Spec: https://spec.matrix.org/v1.9/client-server-api/#mspaceparent
+                // Element-Web: https://github.com/matrix-org/matrix-react-sdk/blob/fb1a97be3228ce1341c54a379f8374710b65a0e8/src/utils/space.tsx#L51-L58
+                type: 'm.space.parent',
+                state_key: parentId,
+                content: {
+                    canonical: true,
+                    via: [this.matrixClient.getDomain()],
+                },
+            });
+        }
 
         const room = await this.matrixClient.createRoom(opts);
         const medienhausMetaEvent = {


pages/spacedeck/[[...roomId]].js
```diff
@@ -140,21 +142,51 @@ export default function Spacedeck() {
...
+        };
+
+        const room = await createRoomForSketch()
+            .catch(async (error) => {
+                return matrix.handleRateLimit(error, () => createRoomForSketch())
+                    .catch(error => setErrorMessage(error.message));
+            });
+
+        // Log debug information about current progress
+        logger.debug('Created Room for %s with id %s ' + name, room);
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:163]: I think the `+` is wrong here. Also, when it comes to string interpolation I think I'd always prefer the ["new" version](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) of

```js
logger.debug(`Created room for ${name} with id ${room}`);
```

pages/spacedeck/[[...roomId]].js
```diff
@@ -140,21 +142,51 @@ export default function Spacedeck() {
...
-            body: link,
-        }).catch(() => {
-            setErrorMessage(t('Something went wrong when trying to save the new sketch link'));
-        });
+    async function createSketchRoom(link, name, parent = serviceSpaceId, retries = 0) {
+        // Log debugging information
+        setUserFeedback(t('Syncing {{name}} from server', { name: name }));
+
+        // Create the room with retry handling
+        const createRoomForSketch = async (retries = 1) => {
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:150]: Is the `retries` parameter ever being used? 🤔

pages/spacedeck/[[...roomId]].js
```diff
@@ -140,21 +142,51 @@ export default function Spacedeck() {
...
+            logger.debug('Attempt %d of creating a room for %s', retries, name);
+
+            return await matrix.createRoom(name, false, '', 'invite', 'content', 'spacedeck');
+        };
+
+        const room = await createRoomForSketch()
+            .catch(async (error) => {
+                return matrix.handleRateLimit(error, () => createRoomForSketch())
+                    .catch(error => setErrorMessage(error.message));
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:159]: I have not tested this yet, but I think that the translation for "Something went wrong..." will never show up. Because where do we call `t()` for the error? I can't seem to find it, but maybe you already know what I mean.

Try switching to German and producing an error. I have a feeling it will show the English version of the error message still.

fnwbr (changes requested):
/

pages/spacedeck/[[...roomId]].js
Comment (marcel-klasse) [reply pages/spacedeck/[[...roomId]].js:150]: it is being used for debug logs which is quite useful.

Comment (fnwbr) [reply pages/spacedeck/[[...roomId]].js:150]: But it's always "1", no?

Comment (marcel-klasse) [reply pages/spacedeck/[[...roomId]].js]: 🤦 missed that when refactoring the code. removed it for now.

pages/spacedeck/[[...roomId]].js
```diff
@@ -149,20 +150,48 @@ export default function Spacedeck() {
...
+        const sendMessage = async () => {
+            await matrixClient.sendMessage(room, {
+                msgtype: 'm.text',
+                body: link,
+            });
+        };
+
+        await sendMessage()
+            .catch(async (error) => {
+                return matrix.handleRateLimit(error, () => sendMessage())
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:192]: So basically you've been adding the very same error handling in two different places in the codebase.
Both -- here in /sketch and also in /write -- we add a space child, and we add the very same retry handler, if something goes wrong. Can't we just do that inside `addSpaceChild()` in `Matrix.js` then?

I'd even say we _should_ do so.

And this also counts for `.createRoom()` for example. Why add a retry handler in all the places that create rooms, when you could just do it in `Matrix.js`?

fnwbr (changes requested):

fnwbr:
Adjusted the code to always show all the present Etherpad/Spacedeck entries we already know about, added some visual eye candy to highlight when we're still synchronizing with the server, removed the item names from the feedback message to tone down on visual changes and possible distraction, and streamlined the code between `spacedeck/[[...roomId]].js` and `etherpad/[[...roomId]].js` to make them both handle the logic in the same manner.

Screenshot 2023-11-21 at 15 10 36

fnwbr:
`/write` is still messing with me... going to continue working on this tomorrow.

PR #73, opened 2023-10-04
Implement Bookmarking Functionality and UI

aofn:
First, basic implementation of a new Bookmark feature, allowing users to add a page or room to their bookmarks and quickly access them from the dashboard or remove them.

- `AddBookmark` component has been introduced, allowing users to add a page or room to their bookmarks from /etherpad or /spacedeck.
- `DisplayBookmarks` component has been introduced, which displays a user's bookmarks and sorts them by service/application
- `Bookmark` component has been introduced to represent a single bookmark entry, displaying the bookmark's name and bin icon to remove the bookmark.

**How bookmarks are stored:**

At the moment, I've decided to save bookmarks inside a bookmark space within the application space.

- Applications
  - spacedeck
  - etherpad
  - bookmarks
    - spacedeck
      - my sketch
    - explore
      - meeting pad
 
This makes it somewhat user friendly within chat clients and also makes it easy for us to tell where a bookmark came from. 

Previously I had stored them as rooms dircetly within the bookmark folder with the title indicating the origin:

- Applications
  - spacedeck
  - etherpad
  - bookmarks
    - my sketch from spacedeck
    - meeting pad from explore

There are multiple ways of doing this, and all have their pros and cons. I've not found a solution so far which strikes me as the absolute best. So I'm very open to suggestions 🙂.
We could also add the actual space/room we are bookmarking to the bookmark structure, similar to example one. 
_I can't remember why I decided not to do this. I have a feeling it had something to do with power levels of the room, but I will update once I remember_

**Important Note:**

It's important to note that checking bookmarks for validity and automating this process would require additional logic. The complexity of this varies depending on the origin of the bookmarked link. For now, users should manually review and remove bookmarks that are no longer needed or valid.


aofn:
@andirueckel I think we have to adjust the design of the component again, since there have been so many changes to `main` in the meantime.

components/UI/ServiceTable.js
```diff
@@ -67,7 +67,7 @@ const TableHeader = styled.th`
   line-height: calc(var(--margin) * 3);
   text-align: ${props => props.align && props.align};
   white-space: nowrap;
-
+  ${props => props.width && `width: ${ typeof props.width === 'string' ? props.width : '8ch'};`}
```
fnwbr, line 70: Hmm... the `<ServiceTable` is supposed to be a super generic and everywhere-to-be-used component, right?
Correct me if I am wrong, but with this change we'd tell every table header to be `8ch` wide, unless we specify a custom width, no?

And all the new usages of `ServiceTable.Header` as of this PR are providing a custom width. So, why even bother?

fnwbr (changes requested):
Also: Missing translations

pages/dashboard/DisplayBookmarks.js
```diff
@@ -0,0 +1,96 @@
...
+ * origin only works
+ */
+const Bookmark = ({ parent, roomId, link, name, origin, removeBookmark }) => {
+    const [removingBookmark, setRemovingBookmark] = useState(false);
+    const { t } = useTranslation();
+
+    const handleRemove = async (e) => {
+        e.preventDefault();
+        setRemovingBookmark(true);
+        await removeBookmark(parent, roomId)
```
fnwbr, line 72: If we're passing the `parent` into the `<Bookmark` component, only to then call the `removeBookmark()` function that is defined in the actual "parent" component `<DisplayBookmarks` we can definitely save up on that.

Instead you can strip away both: `parent` and `removeBookmark` from this `<Bookmark>` component.

And above, in line 45 we can do:
```jsx
    removeBookmark={() => removeBookmark(roomId)}
```

Line 29 can be
```jsx
    const removeBookmark = async (roomId) => {
        await auth.getAuthenticationProvider('matrix').removeSpaceChild(bookmarkSpaceId, roomId);
        return matrix.leaveRoom(roomId);
    };
```

robertschnuell:
Just to wrap up the discussion about bookmarks. There are 4 different approached to specify how to implement the functionality of a bookmark. In any of those cases a bookmark acts quite similar to a ‘quote’ within the matrix, as we take it out of its original context it appeared in and want to reuse it for an user specific purpose in another context. Even if it is just to save it as a bookmark as functionality to favourite any given content. 
So in that case bookmarks is an application in the applications space. 
If a user would like to save something as a bookmark this can be achieved in 3 different ways in that case. (For all of this different approaches it is described in a simplified way that the matrixId would be added as a child directly to the bookmarks matrix room. In practice there would be different type of substructures like the name of the originated applications and so on): 
1. create a new matrix room add the matrix id which should be bookmarked as a child and add the matrix room itself as a child to the bookmarks application space. In this case the origin matrixId would be wrapped into an additional matrix room. 
2. Create a new matrix room and add the https:// link of the medienhaus/spaces url to the newly created matrix room to its timeline. In this case this is a ‘dumb’ link which can’t be linked to its matrix origin itself automatically. 
3. Just add the matrixId which needs to be bookmarked as a child to the bookmarks matrix room.s

Following concerns appeared in the discussion about those candidates: 
- In the 2. approach the saved data would not be interoperable anymore and specifically bound to this specific application. So in that case a federated function would not work anymore as well as we would end up with the same mess which mastodon has to deal with that it is not possible to identify within the application itself if there is a protocol specific data underneath as it is just a link to one specific instantiation of a specific software. It would work for a single purpose use case but would loose all of the benefits, which got created when using matrix as a unified federated database layer.
- In the 3. approach the data would be accessible within matrix but it would not possible to add or modify the bookmark without distorting the origin it came from. In that case if one user wanted to rename a bookmark it would also rename the matrix room in its origin, as long as the user would have also those rights in the original context. So in that case a user specific individualisation would not be possible. 
- the 1. Approach is the most future proof as it would be possible to modify the matrix id which acts as a wrapper based on the user needs. But it would also take the most time to implement. It would be even possible to add bookmark specific data to a bookmark without distorting its original context, like adding a priority or changing the name or adding additional ppl to this bookmark, even if those do not have write rights to the origin of the matrix room which is bookmarked. This tool would also have the option to be used as a tool to achieve the functionality ‘quote’ without write a specification/implementation from ground up. As well as it would still be possible to store ‘dumb’ links from the www. within beside the citation feature it would have within a fully matrix system.


As time is running short, all of those 3 approaches would need more time to implement. So we will take a shortcut and will just store the links to bookmark for now in the account data of the loggedIn user. Still the discussion about a ‘real’ implementation of a bookmark/quote feature was quite valuable and can be continued at some given point in the future from this point on.

aofn:
closed in favour of #131

PR #75, opened 2023-10-05
fix: vertically align service buttons on `/etherpad` and `/spacedeck`

andirueckel:
demo


fnwbr (commented):
Additionally, I think we can remove this stuff now, right?
https://github.com/medienhaus/medienhaus-spaces/blob/6a37b8fe24a0305e834150c86fa380a5369ea694/components/layouts/iframe.js#L80-L87

(... which might have been from me back in the days, totally ignoring the best practice of styled components ...)

fnwbr (approved):

PR #80, opened 2023-10-24
Check if pads are private

marcel-klasse:
This pull request introduces a new feature which uses the MyPads API to check if an etherpad pad is password protected even if they arent't part of their own pads.


pages/etherpad/[[...roomId]].js
```diff
@@ -23,6 +23,34 @@ import CreateAuthoredPad from './actions/CreateAuthoredPad';
 import CreatePasswordPad from './actions/CreatePasswordPad';
 import { path as etherpadPath } from '../../lib/Etherpad';
 
+const WritePads = ({ serverPads, name, writeRoomId, etherpadId, ref, selected }) => {
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:26]: `WritePads` is a bit misleading, when this is only redering exactly **one** link for **one** pad.

pages/etherpad/[[...roomId]].js
```diff
@@ -208,11 +236,12 @@ export default function Etherpad() {
...
                                     if (!name) return;
 
-                                    return <ServiceLink
+                                    return <WritePads
                                         key={writeRoomId}
-                                        name={_.get(matrix.rooms.get(writeRoomId), 'name')}
+                                        name={name}
                                         href={`${etherpadPath}/${writeRoomId}`}
-                                        passwordProtected={serverPads[matrix.roomContents.get(writeRoomId)?.body.substring(matrix.roomContents.get(writeRoomId)?.body.lastIndexOf('/') + 1)]?.visibility === 'private'}
+                                        serverPads={serverPads}
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:243]: This also looks a bit weird.
So we're inside a `.map()` here, where we cycle through _each_ pad that we want to show.

And now -- for some reason -- that single pad needs to have **all the server pads** data structure? That's usually a hint that something can be improved...

fnwbr:
Also, I hear what you're saying about not knowing how to solve this _without_ adding this new component called `WritePads`. It's because you want to run a potentially `async` call for -each- pad in the list. It's `async` if the pad is not our own, as in: if it's not inside of `serverPads`...

I'll help you out with that once the above things are fixed. Feel free to leave these two components separate as they are.

PR #91, opened 2023-11-02
Display login prompt when service token is invalid

andirueckel:
### Changes made

- Added 'use-immer' for improved state management.
- Introduced a 'connectionStatus' object to manage service connection statuses.
- Implemented 'makeRequest' methods within auth providers to handle specific response statuses (401 and 403).
- Modified the 'LoginPrompt' component to display an error message and password `form` when the service connection status is 'false'.

### Testing

- login to spaces
- invalidate your `etherpad` and/or `spacedeck` tokens by for example manipulating them directly in the account data object 
- navigate to /write or /spacdeck
- login prompt should be displayed
- enter password and confirm

Screen Shot 2023-11-22 at 14 45 51


fnwbr:
I like the looks and explicitness of this much better, than the other open PR.

However, I don't see the point of the forward-switcharoo. Let's just render the login prompt right in the page we're trying to access. This is nice and simple, without bloating the `/login` page with concerns that it should not care about.

1. Keep all the custom styling inside the `<LoginPrompt>` component
2. On pages like `/write` or `/sketch`, render _only_ this one component in case the token is invalid by returning early:
  ```jsx
  if (!auth.connectionStatus.spacedeck) return <LoginPrompt />;
  ```
3. Remove all the page-forwarding logic; there won't be any forwarding... `/write` will only render the login prompt; that's effectively what the user is seeing right now, but with additional forwarding logic in the code

marcel-klasse:
style needs to be fixed, since it is not part of IframeLayout. 
@andirueckel could you maybe have a look?

pages/spacedeck/[[...roomId]].js
```diff
@@ -209,7 +211,7 @@ export default function Spacedeck() {
                                         key={spacedeckRoomId}
                                         name={room.name}
                                         href={`${spacedeckPath}/${spacedeckRoomId}`}
-                                        selected={roomId === spacedeckRoomId}
+                                        selected={!auth.connectionStatus.spacedeck || roomId === spacedeckRoomId}
```
Comment (fnwbr) [inline pages/spacedeck/[[...roomId]].js:214]: This is not correct. I understand that you're trying to make the link look disabled, however, this attribute is called `selected`. This is confusing to understand later on in life.

Instead we should add a `disabled` attribute to our `<ServiceLink>` component.

pages/login.js
```diff
@@ -55,35 +66,65 @@ export default function Login() {
...
     const changeHomeserver = () => {
         setHomeserver(prompt(`${t('Set another homeserver')}:`, homeserver) ?? homeserver);
     };
 
-    // Automatically focus the username input on pageload
+    // Automatically focus the username input on page load
     useEffect(() => {
-        usernameInput.current.focus();
-    }, []);
+        if (auth.user !== null && !auth.user) usernameInput.current.focus();
```
fnwbr, line 95: What is this change needed for?

pages/login.js
```diff
@@ -55,35 +66,65 @@ export default function Login() {
...
+        } else {
+            await auth.signin(name, password, homeserver).catch(/** @param {MatrixError} error */(error) => {
+                if (error.errcode === 'M_LIMIT_EXCEEDED') {
+                    setErrorMessage(t('{{error}}, please wait {{time}} seconds.', {
+                        error: error.data.error,
+                        time: Math.round(error.data.retry_after_ms / 1000),
+                    }));
+                } else setErrorMessage(error.data?.error);
+            });
+        }
```
fnwbr, line 84: All this looks like another leftover from the old forward-aroo concept. Can you otherwise explain it to me?

lib/auth/SpacedeckAuthProvider.js
```diff
@@ -47,26 +54,24 @@ class SpacedeckAuthProvider {
     }
 
     async signout() {
-        const response = await fetch(this.url + '/api/sessions/current', {
+        return await fetch(this.url + '/api/sessions/current', {
```
fnwbr, line 57: Never `return await`, only `return`. And then the consumer of this method gets to decide if they want to wait for the process to finish or not.

lib/auth/SpacedeckAuthProvider.js
```diff
@@ -23,11 +34,7 @@ class SpacedeckAuthProvider {
         // If the token is not authorized, the Spacedeck API still returns a payload; therefore we have
         // to check if there _is_ any payload, and if this payload _does_not_contain_ an 'error' key.
         // If these contiditions are matched, then we can assume the token to be momentarily valid.
-        if (tokenValidationCall && !tokenValidationCall.error) {
-            return true;
-        } else {
-            return false;
-        }
+        return !!(tokenValidationCall && !tokenValidationCall.error);
```
fnwbr, line 37: No need for the `!!( )` -- the inside is already going to be a `boolean` at all times.

components/UI/ServiceSubmenu.js
```diff
@@ -4,12 +4,14 @@ import { useTranslation } from 'react-i18next';
...
 
+import TextButton from './TextButton';
+
 const Header = styled.header`
   display: grid;
   grid-template-columns: 1fr auto;
 `;
 
-const ToggleButton = styled.button`
+const ToggleButton = styled(TextButton)`
```
fnwbr, line 14: Now that we're using the `TextButton` as our base here, do we really still need to unset things like `background-color` and `border` in the lines below? (19 & 20)
Isn't the point of the `<TextButton>` component to do exactly _that_?

components/UI/LoginPrompt.js
```diff
@@ -0,0 +1,89 @@
...
+                <>
+                    <ErrorMessage>
+                        { t('Your {{service}} session has expired.', { service: service }) }
+                    </ErrorMessage>
+                    <p>
+                        { t('Please sign in again, in order to use {{service}} without any limitations.', { service: service }) }
+                    </p>
+                </>
+            </ServiceStatus>
+            <InputButton onSubmit={loginAgain}>
```
fnwbr, line 75: `<InputButton>` for something that is in reality a `<form>`... why not just `<Form>`?

pages/spacedeck/[[...roomId]].js
Comment (fnwbr) [reply pages/spacedeck/[[...roomId]].js:214]: Other than that I'm asking myself: Do we really still need this?
Why do we care about making a link's color gray if it can never seen under this condition?

If the connection to the Spacedeck server is interrupted, we show the new login prompt component. We will never show the list of Spacedeck spaces in that case.

pages/etherpad/[[...roomId]].js
```diff
@@ -266,6 +288,7 @@ export default function Etherpad() {
                     <>
                         <ServiceSubmenu
                             title={<h2>{ etherpadPath }</h2>}
+                            disabled={!auth.connectionStatus.etherpad}
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:291]: Why is this still needed? If `!auth.connectionStatus.etherpad` is true (so when the MyPads server is down) we will never ever show this submenu anymore anyways, right?!

pages/etherpad/[[...roomId]].js
```diff
@@ -148,6 +149,25 @@ export default function Etherpad() {
         return room;
     }, [auth, matrix, matrixClient, etherpad]);
 
+    /**
+     * Creates an Etherpad and navigates to it based on the provided name, visibility, and password.
+     *
+     * @param {string} name - The name of the Etherpad to create.
+     * @param {string} visibility - The visibility setting for the Etherpad. Expects 'private' or 'public'.
+     * @param {string} password - The password for accessing the Etherpad (if private).
+     * @returns {Promise<boolean|void>} - A Promise that resolves to a 'false' indicating the success of pad creation or navigates to the Etherpad's URL.
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:158]: This I don't get...

> A Promise that resolves to a 'false' indicating the success of pad creation

`false` indicates the successful creation?

pages/etherpad/actions/CreateAuthoredPad.js
```diff
@@ -1,40 +1,20 @@
...
-            setIsLoading(false);
-
-            return;
-        }
-
-        const link = getConfig().publicRuntimeConfig.authProviders.etherpad.baseUrl + '/' + padId;
-        const roomId = await createWriteRoom(link, padName);
-        router.push(`${etherpadPath}/${roomId}`);
-
-        callbackDone && callbackDone();
```
fnwbr, line 35: So these components used to receive a `callbackDone` function, which would be responsible for... closing the ServiceSubmenu after the process completed.
Without having tested this code: Is this still working? If so, can you explain me how?

(this is a rhetorical question of course; I'm pretty sure this is broken now and the service submenu remains open after you create a new password-protected / authored pad.......)

fnwbr (changes requested):

fnwbr:
> style needs to be fixed, since it is not part of IframeLayout. @andirueckel could you maybe have a look?

Fixed in my very last commit before this comment:
image

components/UI/LoginPrompt.js
marcel-klasse: <Form /> is already in use by our component but I've extracted the component in login.js since I just realised it was basically the same code here and in login.js

pages/spacedeck/[[...roomId]].js
Comment (marcel-klasse) [reply pages/spacedeck/[[...roomId]].js:214]: yeah you're right thats now unnecessary with the new view.

pages/etherpad/[[...roomId]].js
Comment (marcel-klasse) [reply pages/etherpad/[[...roomId]].js:158]: oops that's the wrong way around...

pages/etherpad/actions/CreateAuthoredPad.js
marcel-klasse: is it possible that this has been broken for a while? `callbackDone` was never parsed to the component since it's been changed f178891b49deef15aa9ce6c8b8efcc764af4ca3a . Don't know how it would've closed before 🤔

marcel-klasse: ah nvm, just found it 
https://github.com/medienhaus/medienhaus-spaces/blob/61126628f14c2ab1ebedf141146a3143e97c5071/components/UI/ServiceSubmenu.js#L49-L51

pages/login.js
marcel-klasse: ah yeah, thought this was for validating a session if a user is already logged in. but that doesn't happen here...

fnwbr:
@andirueckel We can't really do the `line-height` trick if this ends up making mobile viewports look like this:
image

I've updated the looks of this by basically removing all CSS that was applied to this `<LoginPrompt>` component. Feel free to re-introduce some, but please test for all viewports and screen sizes.

fnwbr (changes requested):
This is shaping up nicely.

However, logging in again doesn't seem to work for Spacedeck.
@marcel-klasse Try this:

1. Login
2. Verify that /sketch is working perfectly fine
3. Now go invalidate your Spacedeck token by modifying account data via Element
4. Refresh /sketch
5. See the login prompt
6. Enter your password and confirm
7. 💥  Try creating a new sketch

It won't work.

marcel-klasse (approved):

fnwbr (changes requested):
/sketch still broken:

1. Login
2. Verify that /sketch is working perfectly fine
3. Now go invalidate your Spacedeck token by modifying account data via Element
4. Refresh /sketch
5. See the login prompt
6. Enter your password and confirm
7. Try creating a new sketch

Maybe related to this:
image

Because this warning does not show up when logging into the app for the very first time.

PR #96, opened 2023-11-15
Animated service submenu

fnwbr:
@andirueckel @marcel-klasse 
Can you give this a look on your computers and tell me what you think?

Check what happens when you open/close the service submenus in /write or /sketch.

image


andirueckel:
The transition is a bit wonky, but in general I like it. This is (or was) already done in other branches, but the _open menu_ button should be replaced with a _close menu_ button after opening the menu. Also, I’d argue it would be nice if we could click anywhere below the last menu item to close the menu?

fnwbr:
Alright, will refine on those points when I find the time. Thanks for the first check. 👀

marcel-klasse:
😙👌

PR #107, opened 2023-11-30
Fix update active auth providers with matrix account data stored tokens

robertschnuell:
This PR introduces updates mostly to the Matrix.js file, primarily focusing on updating the access tokens stored in the matrix user account data keeping in sync with the active authentication providers.In addition, the `addToken` function of the authProviders was renamed to `setToken` for better readability.  
This code is needed in a multi device use case, in which a new login is forced for all auth providers and should not invalid the functionality of the device already in place.


lib/Matrix.js
```diff
@@ -123,10 +123,32 @@ function useMatrixProvider(auth) {
         }
     }, [directMessages, rooms, setDirectMessages, setInvites, setMessages, setRoom, setRoomContents, setRooms, setSpaces, spaces]);
 
+    const assignAccessTokenToAuthProvider = useCallback((eventContent) => {
+        const activeAuthProviders = auth.getActiveAuthenticationProviderNames();
+        Object.entries(eventContent).forEach(([key, value]) => {
+            if (activeAuthProviders.includes(key)) { // we check if the given accessToken from the accountdata is an active authProvider of this current interface
+                auth.getAuthenticationProvider(key).setToken(value);
+                auth.updateConnectionStatus(key, true); // we assume that this new accessToken is valid and set the connection status to true
```
fnwbr, line 131: Nice 👌

fnwbr (approved):

PR #108, opened 2023-11-30
feat: introduce styled Button…, Input, and Select components

andirueckel:
Introducing styled button, input, and select components for various use-cases, inluding dev route for testing via [`/dev/buttons`](http://localhost/dev/buttons) — this is literally linking to `localhost`, assuming your [`medienhaus-docker-dev`](https://github.com/medienhaus/medienhaus-docker-dev) composition is running locally on your machines.

- [x] `<Button />` with layout and sizing attributes; no color attributes
- [x] `<ButtonPrimary />` with color attributes for primary buttons
- [x] `<ButtonSecondary />` with color attributes for secondary buttons
- [x] `<ButtonInline />` with color attributes for inlined text and/or icon buttons 
- [x] `<Input />`
- [x] `<Select />`

_PS: GIFs and/or images below are slightly outdated by now, but still visually valid in general._

default

Additionally, we have `:hover` styles.

hover-light

hover-dark

And, as wished for, we have `:focus` styles.

focus-light

focus-dark

<del>PS: Still needs to be implemented everywhere™, replacing all _so far, so good_ used `<button>` et cetera elements.</del>


assets/_globalCss.css
```diff
@@ -203,3 +209,4 @@ input[type='password'][disabled] {
   cursor: not-allowed;
   border-color: var(--color-disabled);
 }
+*/
```
fnwbr, line 212: Can you actually remove the old code once you submit this PR as a non-draft one?

components/UI/Button.js
```diff
@@ -0,0 +1,32 @@
+import styled from 'styled-components';
+
+const Button = styled.button`
```
fnwbr, line 3: Looking at everything below: Is this supposed to be a component not actually consumed inside pages, but only act as a "base component" for all the other buttons? (`ButtonPrimary`, `ButtonInline`, ...)
If so, I'd suggest we
- create `/components/UI/buttons/`,
- move this component there, call the component `BaseButton` and call the file `_base.js`,
- move all other buttons also into that directory (and make them all extend that "BaseButton" instead of just "Button")

All of these points are to highlight that the component itself is most likely not to be used or consumed anywhere -- and that someone would most likely want to use -one- of the derivatives instead.

components/UI/Button.js
```diff
@@ -0,0 +1,32 @@
...
+  font: inherit;
+  text-transform: uppercase;
+  cursor: pointer;
+  border-style: solid;
+  border-width: var(--border-width);
+  border-radius: var(--border-radius);
+  appearance: none;
+
+  svg {
+    fill: var(--color-background) !important;
```
fnwbr, line 17: Something feels wrong about the `!important`...
Is there no way of avoiding this?

fnwbr: When taking all of the `!important`s away I end up with this:
image

... which would be fine if we were to discourage using just the `<Button>` component.

fnwbr (commented):
Some more thoughts:

- Why do we need to support icons outside of the `<Icon>` component? I see that with each button you have used both: an icon just by itself (only the `<svg>`), and also an icon inside of our `<Icon>` wrapper...
- What's the `ButtonInlineIcon` component for? Can you add an example to that `dev/` page?

components/UI/Button.js
andirueckel: Hm, yes, we could re-structure the files and re-factor to components this way.

components/UI/Button.js
andirueckel: At some point during development and testing, I had problems with this line: https://github.com/medienhaus/medienhaus-spaces/blob/main/components/UI/Icon.js#L8

Hence, the `!important`; I don’t think I really consider it harmful within a styled component, though we should remove it, if everything works as intended. Maybe the reason why I initially added it is not even a convern anymore after the plenty last commits. ✨

components/UI/Button.js
aofn: This would've been my suggestion as well, like they do it in this example: https://styled-components.com/docs/basics#extending-styles 
I would still prefer the naming scheme how it is atm. So `<Button />` is our base button which all other variants extend on. I can also create a PR for this if you want @andirueckel

PR #112, opened 2023-11-30
Create `m.space.parent` when creating new rooms (with parent)

aofn:
adds [m.space.parent](https://spec.matrix.org/v1.9/client-server-api/#mspaceparent) event when a parent room id is parsed.
`canocical` is set to true, since rooms are newly created, therfore we can assume the parsed parent id is the is the primary parent.

### DIFF
--- a/lib/auth/MatrixAuthProvider.js
+++ b/lib/auth/MatrixAuthProvider.js
@@ -133,6 +133,8 @@ class MatrixAuthProvider {
                 content: { join_rule: joinRule }, // can be set to either public, invite or knock
             },
             {
+                // Spec: https://spec.matrix.org/v1.9/client-server-api/#mspaceparent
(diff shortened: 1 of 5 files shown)


lib/auth/MatrixAuthProvider.js
```diff
@@ -118,7 +131,16 @@ class MatrixAuthProvider {
             {
                 type: 'm.room.join_rules',
                 content: { join_rule: joinRule }, // can be set to either public, invite or knock
-            }],
+            },
+            {
+                type: 'm.space.parent',
```
fnwbr, line 136: I think it's important to now note that: we will always add this.
When calling `createRoom` without a parent, we will still add this `m.space.parent` with an empty `state_key`, right?

I guess this is fine right now. But important to remember when we might want to create rooms that do -not- belong to a parent.

fnwbr (approved):

PR #113, opened 2023-11-30
Knock feature: add UI to knock on spaces with the join rule 'knock'

aofn:
- [x] check if spaces have the joine rule `knock`
- [x] add component to let users request access to the space
- [x] add component to answer knocks #114
- [x] remember and display knocking status (optional)

A door icon is displayed in the explore header, if a space has the join rule `knock` or `knock_restricted`. Not yet the most obvious icon for asking permission for a room, so very open to suggestions.


components/UI/CopyToClipboard.js
```diff
@@ -12,8 +13,7 @@ const CopyToClipboard = ({ content, title }) => {
     const copyToClipboard = async () => {
         navigator.clipboard.writeText(content);
         setWasContentCopied(true);
-        await new Promise(r => setTimeout(r, 2000));
-        setWasContentCopied(false);
+        await waitFor(() => setWasContentCopied(false));
```
fnwbr, line 16: Actually, when I look at all the use cases of `waitFor()` maybe I do want to get back to my comment about `_.delay` -- why do we actually `await` here in this case?

What I'm saying is: for all use cases here I think we can just use `_.delay`. I'll give two examples:

**First example**

This here would just be:
```js
_.delay(() => setWasContentCopied(false), 2500);
```

No need for `await` as far as I can tell.

Which also means we can remove the `async` from `const copyToClipboard = async () => {`.

components/UI/InviteUsersToMatrixRoom/index.js
```diff
@@ -102,10 +103,10 @@ export const InviteUserToMatrixRoom = ({ roomId, onSuccess }) => {
 
         // if everything is okay, we let the user know and exit the view.
         successAmount > 0 && setUserFeedback(<Trans t={t} i18nKey="invitedUser" count={successAmount}>{ { successAmount } } user was invited and needs to accept your invitation</Trans>);
-        await new Promise(() => setTimeout(() => {
+        await waitFor(() => {
```
fnwbr, line 106: **Second example**

This here would just be:
```js
_.delay(() => {
    clearInputs();
    if (onSuccess && successAmount === selectedUsers.length) onSuccess();
}, 3000);
```

fnwbr: ... and so on.
I think the same thing works for all instances of `waitFor()` below. Correct me if I am wrong though.

fnwbr (changes requested):

components/UI/InviteUsersToMatrixRoom/index.js
aofn: yeah looks like it does exactly the same, i'll change them

fnwbr (changes requested):
image

1. It does not make sense to ever show "Invite" and "Knock" at the same time. If it's a room we can knock on, then we cannot invite users. If it's a room we can invite users to, then there's no point in knocking.
2. The "knock" door also shows up for rooms that we are a part of already.

PR #114, opened 2023-12-05
Accept or decline requests to join matrix rooms

aofn:
Accept or reject users requesting access to a Matrix room by _knocking_ on the room:

image


lib/Matrix.js
```diff
@@ -44,6 +45,19 @@ function useMatrixProvider(auth) {
 
         const roomState = room.getLiveTimeline().getState(EventTimeline.FORWARDS);
 
+        // check if any users have knocked on the room
+        const knockingMembers = room.getMembersWithMembership('knock');
```
fnwbr, line 49: This is kind of a bad place for this logic, because now we're just shuffling it down somewhere inside of `buildRoomObject`. It's bad for multiple reasons:

1. `buildRoomObject` is now doing more than what its name says it does; it's updating our internal state, so this becomes an undocumented (and imo unwanted) side-effect
2. At which times exactly do we call this function `buildRoomObject()`? I'd say we don't really know from the top of our heads; so by throwing in the logic to keep our pending knocks up to date, it looks more like: "ah, whatever... this seems to kind of work, let's just put it down somewhere in here"... this leads to:
3. Let's think -when- exactly Matrix' sync call tells us about changes in knocks. I think that it's always a `Room.timeline` event... so why don't we add the logic in there?
4. (Most likely every `Room.timeline` event / each `RoomTimelineEvent()` call is going to be for -one- pending knock?! I'm just guessing here... but this might mean that it should be fairly easy to just take whatever info is inside `event` and push it into our state via `setKnockingMembers`...)

pages/dashboard/KnockCard.js
```diff
@@ -0,0 +1,69 @@
...
+ * @param {string} userId - The ID of the user who made the knock request.
+ * @returns {JSX.Element} A form that allows the user to accept or decline the knock request.
+ */
+export default function KnockCard({ roomId, roomName, user, userId }) {
+    const auth = useAuth();
+    const { t } = useTranslation('dashboard');
+    const matrixClient = auth.getAuthenticationProvider('matrix').getMatrixClient();
+
+    const [isAcceptingKnock, setIsAcceptingKnock] = useState(false);
+    const [isDecliningKnock, setIsDecliningKnock] = useState(false);
```
fnwbr, line 28: Can we combine these two into something like `isProcessingAction`? Or do we actually need the differentiation?

pages/dashboard/index.js
```diff
@@ -118,6 +120,27 @@ export default function Dashboard() {
                     }) }
                 </>
             }
+
+            { pendingKnocks.size > 0 &&
+                <>
+                    <h3>{ t('Accept Knocks') }</h3>
```
fnwbr, line 126: Missing translation

pages/dashboard/KnockCard.js
```diff
@@ -0,0 +1,69 @@
...
+        await matrixClient.kick(roomId, userId);
+        setIsDecliningKnock(false);
+    };
+
+    const handleAccept = async (e, roomId) => {
+        e.preventDefault();
+        setIsAcceptingKnock(true);
+        await matrixClient.invite(roomId, userId);
+        setIsAcceptingKnock(false);
+    };
```
fnwbr, line 42: We can reduce the amount of code here by allowing for a parameter `accept` which can be `true|false` -- and then calling the function below like
```js
        <Form
            onSubmit={(e) => handle(e, roomId, true)}
            onReset={(e) => handle(e, roomId, false)}
        >
```

... and doing something like this here:
```js
    const handle = async (e, roomId, accept) => {
        e.preventDefault();
        setIsAcceptingKnock(true);
        if (accept) {
            await matrixClient.invite(roomId, userId);
        } else {
            await matrixClient.kick(roomId, userId);
        }
        setIsAcceptingKnock(false);
    };
```

(This obviously only works assuming that we can combine the two states as mentioned in my comment above)

fnwbr (changes requested):

components/layouts/partials/navigation.js
```diff
@@ -53,7 +53,7 @@ export default function Navigation({ closeNavigation }) {
     return (
         <>
             <List>
-                <li><Link href="/dashboard">/dashboard{ matrix.invites.size > 0 && <NotificationBubble /> }</Link></li>
+                <li><Link href="/dashboard">/dashboard{ (matrix.invites.size > 0 || matrix.knockedMembers.size > 0) && <NotificationBubble /> }</Link></li>
```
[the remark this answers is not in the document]
aofn: do you mean beacuse of forgetting to change the variable name or for another reason? fixed the name

PR #122, opened 2024-01-10
Persist language selection

aofn:
### Summary 
This PR introduces persistent language handling to the component, saving the default language to localStorage.

### Changes Made
  - Implemented the `changeLanguage` function to save the default language to localStorage and update the language using i18n.
  - Introduced the `preferredLanguage` function to handle preferred language selection based on browser language or localStorage.

### DIFF
--- a/components/layouts/partials/languageChooser.js
+++ b/components/layouts/partials/languageChooser.js
@@ -29,6 +29,11 @@ const LanguageSelect = styled.select`
 function LanguageChooser() {
     const { i18n } = useTranslation();
 
+    const changeLanguage = useCallback((lang) => {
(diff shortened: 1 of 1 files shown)


fnwbr (changes requested):
We're using this npm package for language detection: https://github.com/i18next/i18next-browser-languageDetector

They have a baked-in support for caching & remembering the user choice, which looks like we've just never enabled.

See how we set an empty array `[]` in `Internationalization.js` and see what the plugin allows here: https://github.com/i18next/i18next-browser-languageDetector#detector-options

I think just setting `caches: ['localStorage']` will have the exact same effect as your PR.

fnwbr:
Resolved via 20ace7b6f05aed2ec42f3da4a727ab8585b23021

PR #126, opened 2024-01-19
Tldraw

robertschnuell:


aofn:
looks like draw rooms have the wrong `template`, since they're not being added to the tldraw folder. Haven't check if this is the actual reason, tho, but it's very likely.

PR #131, opened 2024-01-23
bookmark feature using matrix account data

aofn:
Adaptation of #73 using matrix account data to store room id's of spaces.


fnwbr (changes requested):

pages/dashboard/DisplayBookmarks.js
```diff
@@ -0,0 +1,98 @@
...
+ * This component displays a list of bookmarks, stored in the matrix account data, allowing users to remove them.
+ *
+ * @component
+ * @param {string} bookmarkSpaceId - The ID of the space containing bookmarks.
+ * @param {Object} bookmarkObject - The object containing the bookmark data from the matrix account data.
+ * @param {Function} handleRemoveBookmark - A function to remove the bookmark.
+ * @returns {JSX.Element|null} - A JSX element containing the list of bookmarks, or null if the space is not found.
+ */
+
+export default function DisplayBookmarks({ bookmarkObject, handleRemoveBookmark }) {
```
fnwbr, line 41: Can we get rid of this intermediary component?
And just render `<Bookmark>` components right in `dashboard/index.js`?

You can move the logic of calling `getOrigin()` into the `<Bookmark>` component starting in line 67 below instead.

Or is there any benefit of this `<DisplayBookmarks>` component that I'm missing?
(Aside from the bad naming again; it's responsible for **one single** bookmark; so it'd be `DisplayBookmark` instead...)

pages/dashboard/index.js
```diff
@@ -93,6 +95,21 @@ export default function Dashboard() {
         };
     }, [matrixClient, matrix, invitations, setInvitations]);
 
+    const handleRemoveBookmark = async (roomId) => {
```
fnwbr, line 98: Given this function is -only- being used within the `<Bookmark>` component, it should live there instead.

I suppose that the Immer-managed `matrix.bookmarks` variable will be updated by Immer.

fnwbr: When doing this *and* removing `<DisplayBookmarks>` we're going to save up on two pass-downs of this function.

fnwbr (changes requested):

pages/dashboard/index.js
aofn: but wouldn't we instead create a new functions with every render of `DisplayBookmarks`? so for every single bookmark? (same with `getOrigin()`

fnwbr: I know what you mean, from the "memory" point of view you might be correct. But even if so, I think I'd prefer developer experience over a few byte of memory usage in the client.

And by developer experience I mean: have the function as close as possible to where it's being used.

Unless it's being shared by multiple components: in that case it'd make sense to "pull it out" into a parent component of some sort.

But again: In this case this function is only being used in one specific place. So put it as close as possible to that place.

PR #132, opened 2024-01-23
Implement shadcn/ui component library

aofn:
First draft to implement [shadcn/ui](https://ui.shadcn.com/) library and use some of its components.

**shadcn**-related:

- [x] fix `eslint` config (alias/resolver needed?) in `package.json`
- [x] check `components.json` file (see [shadcn/ui docs](https://ui.shadcn.com/docs/components-json))
- [x] `npx`-created components are prone to linting errors due to different default indentation
- [x] can `eslint-import-resolver-alias` be removed (by maybe using `.eslintrc`)?
- [x] ~rename `.jsx` components to `.js`?~ (no)
- [x] ~if rename to `.js`: `@`-imports need literal `.js` file extension in path (see [nextjs docs](https://nextjs.org/docs/pages/building-your-application/configuring/absolute-imports-and-module-aliases))~
- [x] replace `@`-imports with relative paths like `../`, `../../`, etc., or the other way around? (comes via pull-request)

**tailwind**-related:

- [x] iterate over `tailwind.config.js` and remove everything not necessary
- [x] ~maybe re-enable tailwind preflight (see further below)~ (seemingly not necessary)
- [x] decide if we define css variables in `_globalCss.css` or `tailwind.config.js`
- [x] define all necessary css variables to make tailwind happy
- [x] ~do we need a separate `tailwind.css` file? (see [shadcn/ui docs](https://ui.shadcn.com/docs/components-json#tailwindcss))~ (no)
- [x] do we want to use a prefix for tailwind classes (`tw-`), or should we rather remove it? (see [shadcn/ui docs](https://ui.shadcn.com/docs/components-json#tailwindprefix)) (removed)

~`card.jsx` was customised by removing the `shadow-sm` box-shadow class and replacing it with `border-solid` class. box-shadow is not displayed for some reason.~

[tailwind preflight](https://tailwindcss.com/docs/preflight) is currently deactivated.

we need to decide which route we want to go, either deactive it and adapt components where necessary (since they are built upon preflight styles) or activate preflight and adapt our css and/or html where necessary.

You can activate preflight inside `tailwind.config.js`

@fnwbr 
@andirueckel


andirueckel:
Good first draft! 🙂

aofn:
image
instead of alerts maybe?

andirueckel:
> instead of alerts maybe?

 What about using the same positioning and rounded corners for these kinds of styled *inline alerts* as for the `Connection lost` message? This could be a new shadcn/ui based styled component that is being used for both (and maybe soon more) occasions.

fnwbr:
Regarding replacing `alert` or `confirm`: Only as long as the replacement (e.g. the toast) is tested and working well on
- every client platform (Linux, macOS, Windows...)
- every browser (Firefox, Chrome, Safari), and
- every screen size (desktops, tablets, smartphones)

Given that is quite the amount of work, I'd suggest to not replace them right now.

aofn:
should we change `CardTitle` in `card.jsx`  from:

```
const CardTitle = React.forwardRef(({ className, ...props }, ref) => (
    <h3
        ref={ref}
        className={cn(
            'tw-text-2xl tw-font-semibold tw-leading-none tw-tracking-tight',
            className,
        )}
        {...props} />
));
```
to

```
const CardTitle = React.forwardRef(({ className, children, ...props }, ref) => (
    <h3
        ref={ref}
        className={cn(
            'tw-text-2xl tw-font-semibold tw-leading-none tw-tracking-tight',
            className,
        )}
        {...props}>
{children}
</h3>
));
```

to avoid the following eslint error:

> ESLint: Headings must have content and the content must be accessible by a screen reader.(jsx-a11y/heading-has-content)

andirueckel:
I [drafted](https://github.com/medienhaus/medienhaus-spaces/pull/136) enabling the Tailwind preflight, removing our own to-be-deprecated global CSs, and adding shadcn/ui `Button` and `Input` components to `/login` and `/dashboard`.

Maybe we can use that as base for further drafting, testing, desicions, and implementation.

fnwbr:
Screenshot 2024-01-31 at 12 32 39

- [ ] Secondary button looks too much like a disabled primary button in my opinion.

---

Screenshot 2024-01-31 at 12 33 53

- [ ] Too little vertical padding, too much horizontal padding.
I think this looks better for example:
image

---

- [ ] What happened to the font weight hierarchy we had? Let's bring it back, please! 🥺

_Before_
image

_After_
Screenshot 2024-01-31 at 12 32 07

andirueckel:
TODO: The opened shadcn/ui quasi `<option>` menu is somewhat wider than the shadcn/ui quasi `<select>` component.

Screenshot 2024-01-31 at 16 57 14

assets/_globalCss.css
```diff
@@ -1,49 +1,84 @@
...
 }
 
 @media (prefers-color-scheme: dark) {
     :root {
-        --color-background: hsl(0deg 0% 8%);
-        --color-foreground: rgb(255 255 255);
-        --color-disabled: rgb(128 128 128);
-        --color-selection: rgb(128 196 255);
-        --color-notification: rgb(96 64 255);
+        --background: 0 0% 0%;
```
fnwbr, line 51: We overlooked something here and accidentally changed `8%` to `0%`. Fixed in https://github.com/medienhaus/medienhaus-spaces/commit/f5bb18d028c6399e00547cdc1f64a421079bd050.

PR #133, opened 2024-01-24
/write: simplify and streamline workflow

### LINKED ISSUE #130: /write: simplify and streamline new item workflow
For our first release candidate, we wanted to simplify and streamline the *new item* workflow for `/write`. Currently there’s too many options, as https://github.com/medienhaus/medienhaus-spaces/issues/129 describes, hence the following suggestion with conditional `<input type="password">` field.

```
/write           [+]    |  h2 + sub-menu button
                        |  br
Name your item:         |  h3
_______________         |  input type="text"
                        |  br
[✓] Password?           |  input type="checkbox" + h3
_______________         |  input type="password"
```

Additionally, *MyPads* allows to **change the *visibility*** from public (world-readable if URL is known) to private (password-protected), and vice-versa; as well as **changing the password** itself. In my opinion, adding this functionality would be more important than adding external public pads via URL—which should be handled similar to other hyperlinks, not */write*-specific.

What do you think, @aofn @fnwbr?

aofn:
closes #130

### DIFF
--- a/pages/etherpad/[[...roomId]].js
+++ b/pages/etherpad/[[...roomId]].js
@@ -169,7 +169,8 @@ export default function Etherpad() {
      */
 
     const createPadAndOpen = async (name, visibility, password) => {
-        const padObject = await etherpad.createPad(name, visibility, password);
+        // truncate the name to a maximum of 40 characters to avoid errors
+        const padObject = await etherpad.createPad(name.substring(0, 40), visibility, password);
         if (!padObject || !padObject.success) throw new Error('The following error occurred', { cause: padObject });
 
         const link = getConfig().publicRuntimeConfig.authProviders.etherpad.baseUrl + '/' + padObject.key;


fnwbr:
🐛 Assuming that this PR is the culprit: on our current testing environment the "Create new pad" panel does **not** automatically close anymore, after someone clicked "Create pad". Instead it stays open.

Also, I'm missing some whitespace here:
Screenshot 2024-02-07 at 14 14 37

andirueckel:
@fnwbr Voilà, whitespace.

Maybe the *animated service submenu* approach would be even better? https://github.com/medienhaus/medienhaus-spaces/pull/96

Screenshot 2024-03-06 at 17 33 39

pages/etherpad/actions/CreateNewPad.js
```diff
@@ -0,0 +1,89 @@
...
+        });
+        setIsLoading(false);
+        if (!createPad) return;
+        callbackDone && callbackDone();
+        setPadName('');
+    };
+
+    return (
+        <form
+            className="mb-8 [&>*+*]:mt-4"
```
fnwbr, line 50: Instead of using this selector, how about adding a `mt-4` to every component below? I think that would be the more correct "Tailwind"-approach.

pages/etherpad/[[...roomId]].js
```diff
@@ -25,6 +20,16 @@ import logger from '@/lib/Logging';
 import { isMyPadsApiEnabled, path as etherpadPath } from '@/lib/Etherpad';
 import { useMatrix } from '@/lib/Matrix';
 import { useAuth } from '@/lib/Auth';
+import CreateNewPad from './actions/CreateNewPad';
+
+const Header = styled.header`
```
Comment (fnwbr) [inline pages/etherpad/[[...roomId]].js:25]: Can we use Tailwind instead of making these two new styled components?

fnwbr (changes requested):
The new component feels very imbalanced in regards to whitespacing:
image

It also breaks with what the other pages do. I'm okay with changing it if the both of you prefer the more air-y, more whitespace-y version, but let's do so consistently then.

fnwbr:
Also, when the MyPads API is **not** enabled in the config, we still show the user the password inputs:
image

Clicking on "Create pad" errors:
image

PR #139, opened 2024-02-02
accessing the header actions of element iframe via parent

robertschnuell:
To achieve a more consistent UI, it is necessary to include the same header that we implemented on all other routes, also in the /chat route. Unfortunately, Element brings its own header. We eliminate this by hiding it and accessing the functionality of those icons via the main header, by propagating the click on the icon into the iframe. This is only possible as long as the iframe URL is deployed on the same domain as the medienhaus/spaces deployment itself.


fnwbr (changes requested):

pages/chat/[[...roomId]].js
```diff
@@ -195,15 +207,85 @@ export default function Chat() {
...
             </DefaultLayout.Sidebar>
             {roomId && (
                 <DefaultLayout.IframeWrapper>
-                    <iframe
-                        ref={iframe}
-                        title="/chat"
-                        src={`${getConfig().publicRuntimeConfig.chat.pathToElement}/#/${roomId === 'new' ? 'home' : `room/${roomId}`}`}
-                    />
+                    <DefaultLayout.IframeHeader>
+                        <h2>{matrix.rooms.get(roomId).name}</h2>
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:214]: 💥 This bricks when clicking the "new chat" button...

fnwbr (changes requested):

fnwbr (changes requested):

pages/chat/[[...roomId]].js
```diff
@@ -195,10 +204,69 @@ export default function Chat() {
...
+                                                .querySelector('header.mx_RoomHeader > div button:nth-child(1)')
+                                                .click()
+                                        }
+                                    >
+                                        <Icon>
+                                            <RiPhoneLine />
+                                        </Icon>
+                                    </TextButton>
+                                    <TextButton
+                                        variant="ghost"
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:253]: For these `<TextButton>` components there are no "variants"... so this line should be removed.

pages/chat/[[...roomId]].js
```diff
@@ -195,10 +204,69 @@ export default function Chat() {
...
+                                    />
+                                    <CopyToClipboard title={t('Copy chat link to clipboard')} content={'chat/' + roomId} />
+                                    <TextButton variant="ghost" title={t('Leave chat')} onClick={leaveRoom}>
+                                        <Icon>
+                                            <RiDoorOpenLine />
+                                        </Icon>
+                                    </TextButton>
+                                    <TextButton
+                                        variant="ghost"
+                                        title={t('call')}
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:241]: Uppercase "Call" for consistency pls; also: no German translation for this?

pages/chat/[[...roomId]].js
```diff
@@ -195,10 +204,69 @@ export default function Chat() {
...
+                                                variant="ghost"
+                                                title={t('Invite users to {{name}}', { name: matrix.rooms.get(roomId).name })}
+                                            >
+                                                <Icon>
+                                                    <RiUserAddLine />
+                                                </Icon>
+                                            </TextButton>
+                                        }
+                                    />
+                                    <CopyToClipboard title={t('Copy chat link to clipboard')} content={'chat/' + roomId} />
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:233]: This copies `chat/<roomId>` to my clipboard... shouldn't this be a working link that starts with "https://" so the user can send this to someone else?

pages/chat/[[...roomId]].js
Comment (fnwbr) [reply pages/chat/[[...roomId]].js:253]: (same goes for the other places where we're using `<TextButton>` in this PR)

pages/chat/[[...roomId]].js
```diff
@@ -143,6 +151,7 @@ export default function Chat() {
             <DefaultLayout.Sidebar>
                 <h2>
                     <TextButton
+                        variant="ghost"
```
Comment (fnwbr) [inline pages/chat/[[...roomId]].js:154]: can be removed

pages/chat/[[...roomId]].js
Comment (fnwbr) [reply pages/chat/[[...roomId]].js:233]: I guess another option would be to just copy the room ID in the "Matrix style format" of `!3453543tjjrea3123:matrix.org`... at least when pasting this into a Matrix chatroom it would also just work.

Or build a link using the `matrix.to` service, like:
https://matrix.to/#/!hqcBkCRsdsrSZplHmV:medienhaus.dev/$yg6SZ6nsISgSf5kns7WztQiAVIs_myxwyXw1XVPfAtg?via=medienhaus.dev for example.

fnwbr (changes requested):

PR #143, opened 2024-02-14
Add shadcn/ui <DialogDrawer> component; move InviteUsersToMatrix into a <Dialog>

fnwbr:
Requires https://github.com/medienhaus/medienhaus-spaces/pull/132

---

## Preview

 image

PR #144, opened 2024-02-14
/explore: restructure design and workflow of explore settings/actions

aofn:


andirueckel:
Some screenshots of the current but still rather intermediate state.

The settings page, or rather the referenced components, need some layout changes.

We should maybe start implementing shadcn/ui drawer in addition to the already implemented dialog, so we can “open” the members and settings menu in the dialog (on wide screens) and drawer (on narrow screens). It seems that some work on implementing the drawer has already been done; see: https://github.com/medienhaus/medienhaus-spaces/pull/143

And we could discuss when to implement the shadcn/ui data table to replace our own service table component.

Screenshot 2024-02-15 at 15 00 59
Screenshot 2024-02-15 at 15 01 05
Screenshot 2024-02-15 at 15 01 37
Screenshot 2024-02-15 at 15 01 46
Screenshot 2024-02-15 at 15 01 49
Screenshot 2024-02-15 at 15 01 59
Screenshot 2024-02-15 at 15 02 02

PR #149, opened 2024-03-05
/explore: use sonner to display success message for space actions

aofn:
Displays a success message in a sonner component after adding content to a space in /explore


andirueckel (approved):
Added missing locales; everything else looks fine to me.

PR #151, opened 2024-03-12
Onboarding

robertschnuell:
This ~draft~ PR introduces an onboarding feature for new users after their first login. It navigates the user based on the `onboardingScript.json` through different routes. It is based on driver.js and also uses the sheet component of shadcn.

The following functions are still missing:
- [x] Show /intro route only after the first login
 - [x] ~Accepting the terms should be stored server-side (like we did some years ago in the CMS). Separate PR?~ Seperate PR!
- [x] Texts for the intro page
- [x] Text for the onboarding routes and steps
- [x] Refactoring of the sheet component into an independent one, so that it will not permanently manipulate the shadcn one
- [x] Storing the progress of the onboarding in matrix account data, to have cross-device functionality
- [x] Fixing ~a lot of~ some CSS, especially on mobile
- [x] Relocate `onboardingScript.json` so that it can be ignored same way as the `config.js`
- [x] 'jump back to previouse route' functionallity.
- [x] Locales …


components/onboarding/onboardingContext.js
```diff
@@ -70,8 +80,8 @@ function useOnboardingProvider() {
         // start the tour if the instance is created, which is the case after a route change
         if (tourInstance) {
             tourInstance.drive();
-            setCurrentStepDescription(tourInstance.getActiveStep().popover.description);
-            setCurrentStepTitle(tourInstance.getActiveStep().popover.title);
+            setCurrentStepDescription(tourInstance?.getActiveStep()?.popover?.description);
+            setCurrentStepTitle(tourInstance?.getActiveStep()?.popover?.title);
```
aofn, line 84: tourInstance doesn't need a `?`, you're already  in a conditional `if(tourInstance) {...}`

components/onboarding/onboardingPilot.js
```diff
@@ -1,18 +1,40 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect, use } from 'react';
```
aofn, line 1: `use`

pages/intro.js
```diff
@@ -88,7 +91,12 @@ export default function Intro() {
                         disabled={!termsAccepted}
                         variant="onboarding"
                         onClick={() => {
-                            onboading.startTour();
+                            onboarding.startTour();
+                            onboarding.writeOnboardStateToAccountData(matrixClient, {
+                                completed: false,
+                                active: true,
+                                currentRouteIndex: 0,
+                            });
```
aofn, line 111: do we need two states for this? isn't `active` enough?

pages/_app.js
```diff
@@ -9,8 +9,15 @@ import DefaultLayout from '@/components/layouts/default';
...
+
 import '../lib/Internationalization';
 import '../assets/_globalCss.css';
+
+import '../assets/driverJsCustom.css';
+
+import { useOnboardingProvider, OnboardingContext } from '@/components/onboarding/onboardingContext';
+import OnboardingPilot from '@/components/onboarding/onboardingPilot';
+
 import { Toaster } from '@/components/UI/shadcn/Sonner';
```
aofn, line 21: check eslint errors for the file please.

components/onboarding/onboardingContext.js
```diff
@@ -0,0 +1,204 @@
...
+    const [hasPrev, setHasPrev] = useState(false);
+    const [hasNext, setHasNext] = useState(false);
+    const [currentStepDescription, setCurrentStepDescription] = useState('');
+    const [currentStepTitle, setCurrentStepTitle] = useState('');
+
+    const [driverJsConfig, setDriverJsConfig] = useState(null);
+
+    const [tourInstance, setTourInstance] = useState(null);
+
+    const onboadingScript = onboadingScriptCustom?.length > 0 ? onboadingScriptCustom : onboadingScriptGeneric;
```
aofn, line 34: typo: onboading

components/onboarding/onboardingContext.js
```diff
@@ -0,0 +1,204 @@
+import { createContext, useContext, useEffect, useState } from 'react';
+import { useRouter } from 'next/router';
+import { driver } from 'driver.js'; // import driver.js
+
+import 'driver.js/dist/driver.css'; //import css
+import onboadingScriptGeneric from './onboardingScriptGeneric.json';
+import onboadingScriptCustom from './onboardingScriptCustom.json';
```
aofn, line 7: typo: onbaoding

aofn (changes requested):

pages/intro.js
robertschnuell: I would say: yes

robertschnuell: all jokes aside: with the 'completed' key, we can distinguish whether this is the first time a user needs to do the onboarding or if the user would like to revisit one of those steps once more in the future. I will add a function that shows a 'close' button if 'completed' is true.

components/layouts/partials/navigation.js
```diff
@@ -55,33 +63,43 @@ export default function Navigation({ closeNavigation }) {
         <>
             <List>
                 <li>
-                    <Link href="/dashboard">
+                    <Link active={onboarding?.active} href="/dashboard">
```
aofn, line 66: 😙👌
but not working for /chat yet

aofn (changes requested):

components/onboarding/onboardingContext.js
```diff
@@ -0,0 +1,204 @@
...
+    const [currentStepDescription, setCurrentStepDescription] = useState('');
+    const [currentStepTitle, setCurrentStepTitle] = useState('');
+
+    const [driverJsConfig, setDriverJsConfig] = useState(null);
+
+    const [tourInstance, setTourInstance] = useState(null);
+
+    const onboardingScript = onboardingScriptCustom?.length > 0 ? onboardingScriptCustom : onboardingScriptGeneric;
+
+    const isScriptCustom = onboardingScriptCustom?.length > 0;
```
aofn, line 36: We're just exporting this const to know which translation file to use? in that case I would say we can just use the conditional from line 34 instead.

components/onboarding/onboardingContext.js
```diff
@@ -0,0 +1,204 @@
...
+    const [currentRoute, setCurrentRoute] = useState('');
+    const [nextRouteName, setNextRouteName] = useState('');
+
+    const [prevRouteName, setPrevRouteName] = useState('');
+
+    const [currentRouteIndex, setCurrentRouteIndex] = useState(0);
+
+    const [currentSteps, setCurrentSteps] = useState([]);
+
+    const [currentStep, setCurrentStep] = useState(0);
```
aofn, line 24: are we using `currentStep` anywhere? If so, I think we should rename `currentSteps` (plural) to `currentRouteSteps` or something similar, it's hard to distinguish them both when reading the code.
But i don't think `currentStep` is being used.

components/onboarding/onboardingContext.js
```diff
@@ -0,0 +1,204 @@
...
+    const [hasPrev, setHasPrev] = useState(false);
+    const [hasNext, setHasNext] = useState(false);
+    const [currentStepDescription, setCurrentStepDescription] = useState('');
+    const [currentStepTitle, setCurrentStepTitle] = useState('');
```
aofn, line 28: I'm half thinking maybe we should just create one `useImmer` state for this? then we could just have one object like this:
```
const [route, setRoute] = useImmer({
   name: "chat",
   nextName: "account"
   index: 1,
   currentSteps: [...],
   ....
})

then we wouldn't have so many odovidual states, and the consumer of the context could just get `onboarding.route` and have all the informations for the current route

aofn (changes requested):

PR #156, opened 2024-03-13
/explore: restructure header buttons

andirueckel:
@aofn should be prototypically ready for review and merge.

The `Members` view needs to be updated to `DataTable`, but that shall be done after merge in another pull-request.

Screenshot 2024-03-13 at 17 31 15

Screenshot 2024-03-13 at 17 36 08


andirueckel:
> hover colour in dark mode is a bit hard to see imo, but can be changed later (maybe also just underlined?)

The table row highlight hover background color? That I wanted to be rather subtle.

PR #158, opened 2024-03-19
tldraw using /sync

fnwbr:
This is a Work-In-Progress to embed [tldraw](https://tldraw.dev/), based on the prototype https://github.com/medienhaus/medienhaus-spaces/pull/126. It respects the "human-readable" first approach.

---

**Todos:**

- [ ] ‼️ use an interval to send updates to Matrix when dragging/moving shapes, changing text on notes
- [ ] promote users to moderator when inviting them (or toggle switch to decide if they should be able to(read/write))
- [ ] if there is a read-only view: disable all functions if a user is read-only
- [ ] fix the autofocus (aka fix keyboard shortcuts)
- [ ] use self-hosted icons instead of unpkg.com
- [ ] hide all features from the UI that are currently not supported, or support them:
    - [ ] uploading images
    - [ ] using different pages
    - [ ] ... ?
- [ ] various `@TODO` comments in the code


robertschnuell:
'promote users to moderator when inviting them (or toggle switch to decide if they should be able to(read/write))' has been solved in `develop` by @aofn   -> https://github.com/medienhaus/medienhaus-spaces/commit/ff74c4b34c46b8ec4887a80f1593a21ffe73473a

PR #161, opened 2024-04-08
/explore: fix icons in context view

andirueckel:
@robertschnuell Fixing the *wrong icon* part of your reported bug; please test locally, and quickly review the changes.


robertschnuell (approved):
We can leave this suggested changes to align it with IframeSidebar.js, anyhow the problem was actually something else as the link was created with the wrong meta type (`context` instead of `item`)

PR #164, opened 2024-04-16
/explore: conditional loading

andirueckel:
Concerning the following issue from one of our evergrowing etherpads:

> description:
> in branch `explore`, header buttons and tab trigger buttons appear too soon
> 
> how to reproduce:
> navigate to `/explore`, then navigate to some sub-context which you moderate
> 
> problem:
> “new” header buttons and tab trigger buttons appear before “new” title and content are fetched/loaded; that means, the “new” header buttons and tab trigger buttons matching the context we navigated to, but title and content of old context (where we come from) are displayed for some time, which is quite intriguing.

The proposed maybe-workaround displays our loading spinner and progress bar with no content at all—no header buttons, no tab trigger buttons, no title, and no content—until we’re finished fetching the content to be displayed/rendered.

Unfortunately, some times—though not always—, right after making the navigation click to the new context, the tab trigger buttons briefly switch to a different state before nothing but the loading spinner and progress bar are being shown.

Maybe @robertschnuell you feel like debugging this, else we could merge and either deploy the old and rather intriguing behaviour, or the new and a little less intriguing behavior … 👀

Screenshot 2024-04-16 at 15 59 13

### DIFF
--- a/pages/explore/[[...roomId]].js
+++ b/pages/explore/[[...roomId]].js
@@ -403,10 +403,10 @@ export default function Explore() {
                                         {table.getRowModel().rows?.length > 1 && (
                                             <Table>
                                                 {/*
-                                                      @NOTE: we cannot use border-top/-bottom for sticky thead (because borders scroll with the content);
-                                                      fortunately this does not apply to box-shadow, hence the madness below; we also increase the height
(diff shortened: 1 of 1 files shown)


robertschnuell (approved):
works on my machine. would not share the same opinion that the `settings` button always need to be displayed even if there are no options to interact with. But I don't really care either.
approved and merged.

PR #166, opened 2024-06-05
Reintroduce getIcon method to slim code in explore

aofn:
Reintroducing the getIcon method which was removed in 5d366e55f6fd094abfe2c435663f3140261d8d4c and 5284563c7f56a6dfd6ddf3af226331a72c7221c3.  
This makes it easier to add new icons or move the function into its own file in case a lot of icons are added an the funfiton becomes very long.


andirueckel:
Looks good; but the pull-request should probably be made against our `explore` branch?

PR #173, opened 2024-10-24
Ensure upstream compatibility for custom user css changes

aofn:
Adds an import to the global css file and a `users.css` file, where users can customise the style of the app.
Workaround to introducing theme support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,5 +40,3 @@ yarn-error.log*
 # JetBrains IDE related things
 /.idea/workspace.xml
 
-# custom user css
-assets/user.css
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ yarn-error.log*
 
 # JetBrains IDE related things
 /.idea/workspace.xml
+
+# custom user css
+assets/user.css
--- a/README.md
+++ b/README.md
@@ -50,6 +50,6 @@ We use [`ESLint`](https://github.com/eslint/eslint) and [`Prettier`](https://git
 
 ### Customisation
 
-You can add custom styles by defining them in `users.css` file to the `assets` directory. This file will be loaded after the default styles.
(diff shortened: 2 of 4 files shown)


andirueckel:
@aofn I mean, why not; looks fine to me.

Should we rename the file to something like `userstyle.css`, or `overwrites.css`?

What’s the rationale behind `_globalCss.css` being prefixed by an underscore, in case you know?

Would we rather `@import` the new user/userstyle/overwrite CSS file at the very end of `_globalCss.css` file? That *might* reduce the necessity of *by default* having to use the `!important` flag, if I’m not completely mistaken. The `!important` flag might still be necessary in *some* cases, though.

aofn:
yeah we can rename it, don't have a prefernce there. I'm not sure about the underscore, maybe coming from scss? 

I don't think we can add the import at the end, as imports always have to be [at the top of the file](https://developer.mozilla.org/en-US/docs/Web/CSS/@import), so it will create an error while building. I think it also impact performance IIRC.

andirueckel:
> I don't think we can add the import at the end, as imports always have to be [at the top of the file](https://developer.mozilla.org/en-US/docs/Web/CSS/@import), so it will create an error while building. I think it also impact performance IIRC.

Interesting, I might not have been aware of this; how about conditionally — if exists, else don’t; in case that’s feasible — importing the file via `/pages/_app.js`, as we do for `../assets/_globalCss.css` and `../assets/driverJsCustom.css` as well as `tldraw/tldraw.css` and `../assets/_tldrawOverrides.css`?

- https://github.com/medienhaus/medienhaus-spaces/blob/develop/pages/_app.js#L13-L14
- https://github.com/medienhaus/medienhaus-spaces/blob/develop/pages/_app.js#L20-L21

But … in case that would make things more complicated than necessary, please ignore my remarks and _do the merge_.

PR #175, opened 2024-11-05
Conditional loading of custom user css file

aofn:
- Added dynamic import for the CSS file based on a condition.
   - Added `enableUserCssOverrides` as a config flag for CSS overrides by users in `config.js`.
   - Added `customCssPath` as a config flag for CSS overrides by users  in `config.js`.

An alternative approach to having an empty file within our code base.


andirueckel (approved):
👏

PR #177, opened 2024-11-27
Edit room name within explore

aofn:
Adds a component to explore settings to let users change the name of a space/room.

The room name in the header of explore is usually one step behind. This was a problem we encountered and weren't able to find the solution so far. Sometihing within `lib/Matrix.js` doesn't seem to fire when changing the name or is just one step behind.

users are therefore prompted to reload the page in order for changes to take (visual) effect

### DIFF
--- a/pages/explore/TreePath.js
+++ b/pages/explore/TreePath.js
@@ -11,7 +11,7 @@ import {
     BreadcrumbSeparator,
 } from '@/components/UI/shadcn/Breadcrumb';
 
-const TreePath = ({ selectedSpaceChildren }) => {
+const TreePath = ({ selectedSpaceChildren, roomName }) => {
     return (
         <>
             <Breadcrumb className="overflow-hidden">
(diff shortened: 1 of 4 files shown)


robertschnuell:
sadly this PR does not work on my machine. even after a clean install I always get the error: `TypeError: Cannot read properties of undefined (reading 'getMatrixClient')` in file `lib/Matrix.js (10:48) @ getMatrixClient`. private session or clearing all localstorage does not solve the problem either.
any idea why this isnt working?

aofn:
hm just tested this again and it was working on my side. Does everything else work normally? Does auth work fine? Because the error kind of sounds like the matrixClient isn't initialised or something like that, since it's pointing to `lib/Matrix.js`. Does the console give any more detais?

robertschnuell:
seems like your other PR somehow works without any next.js config, but this one does not. anyhow that was the problem that this file went missing. sorry for that, the origin of the problem seemed to be infront of the computer.

PR #179, opened 2024-11-29
display the homeserver next to federated spaces inside /explore

aofn:
This pull request introduces the display of federated spaces within /explore. 
If a space is not from the origin domain, the domain of the homeserver is displayed next to the name of the space with the tailwind variable `text-muted`.


robertschnuell (approved):
works on my machine.

